PR #8, opened 2019-11-25
Dmitry Goman - React Task 1

dima-red:


src/components/blocks/Header/Header.jsx
```diff
@@ -0,0 +1,29 @@
...
+import Logo from 'components/reusable/Logo/Logo';
+import { IconButton } from '@material-ui/core';
+import HomeIcon from '@material-ui/icons/Home';
+import Avatar from '@material-ui/core/Avatar';
+import { header, headerContent, divider } from './Header.scss';
+import { avatar } from '../../reusable/Avatar/Avatar.scss';
+import avatartSrc from 'assets/images/DG.png';
+
+const logoProps = {
+  text: 'D&G (Dima & Goman)'
```
AV29: Hulk & Bruce Banner :)

src/components/blocks/Header/Header.jsx
```diff
@@ -0,0 +1,29 @@
...
+    <header className={ header }>
+      <div className={ headerContent }>
+        <IconButton >
+          <HomeIcon fontSize="large"/>
+        </IconButton>
+        <div className={ divider }></div>
+        <Logo logoContent={ logoProps }/>
+      </div>
+      <div className={ avatar }>
+        <Avatar alt="avatar" src={ avatartSrc } />
```
AV29: Could this Avatar (svg) component take className property? if so - then <div className={ avatar }> is redundant

src/components/blocks/Main/Main.jsx
```diff
@@ -0,0 +1,32 @@
...
+import { mainContent, mainWrapper, cityItemWrapper } from './Main.scss';
+
+const Main = (props) => {
+    const { cityWeatherReducer } = props.cityData;
+
+    return (
+        <main className={ mainWrapper }>
+            <Switch>
+                <Route exact path="/">
+                    <Paper className={ mainContent }>
```
AV29: Component Paper... what does it do ? I personally don't get it ...

src/components/blocks/SideBar/SideBar.scss
```diff
@@ -0,0 +1,17 @@
+.leftGrid {
+    // min-width: 50px;
```
AV29: IMHO, let's not commit comments

src/components/pages/CurrentCity/CurrentCity.jsx
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
+import { Button } from '@material-ui/core';
+import Grid from '@material-ui/core/Grid';
+import { Link } from "react-router-dom";
+import WeatherIcon from '../../reusable/WeatherIcon/WeatherIcon';
```
AV29: this import should be on line 8 instead

src/components/pages/CurrentCity/CurrentCity.jsx
```diff
@@ -0,0 +1,77 @@
...
+    date,
+    time,
+    city,
+    temperature
+} from './CurrentCity.scss';
+
+class CurrentCity extends React.Component {
+    constructor(props) {
+        super(props);
+        this.cityData = this.props.cityData;
```
AV29: the problem you may face with is that when props change - your component (CurrentCity) won't reflect those changes. Constructor executes only once per Component's lifecycle setting props data to **this**. Then you are using instance (this) data do reflect it on a view. You may never notice in your specific ad-hoc case: say, your CurrentCity component being rerendered from the scratch every time executing constructor each time. But this kind of approach is bug prone.

Instead , why not just render data from props directly ?

src/components/pages/CurrentCity/CurrentCity.jsx
```diff
@@ -0,0 +1,77 @@
...
+                    </Grid>
+                </Grid>
+                <div className={ editLocationButtonWrapper }>
+                    <Link to="/cities">
+                        <Button
+                            size="large"
+                            variant="contained"
+                            color="default"
+                            id={ editLocationButton }
+                            startIcon={ <EditLocationIcon fontSize="large"/> }
```
AV29: what does startIcon property mean ? 

Could this Button  and IconButton be merged somehow into one component ?

src/components/reusable/CityItem/CityItem.jsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import Grid from '@material-ui/core/Grid';
+import WeatherIcon from '../../reusable/WeatherIcon/WeatherIcon';
```
AV29: How about move WeatherIcon import on line 6

src/components/reusable/CityItem/CityItem.jsx
```diff
@@ -0,0 +1,55 @@
...
+import { Link } from "react-router-dom";
+import {
```
AV29: how about just **import styles from './CityItem.scss'** ? ...and place it at the end of import block

src/components/reusable/CityItem/CityItem.jsx
```diff
@@ -0,0 +1,55 @@
...
+    windWrapper,
+    gridItem,
+    windText,
+    city,
+    temperature
+} from './CityItem.scss';
+import { setCurrentCity } from '../../../actions/cityActions';
+
+class CityItem extends React.Component {
+    cityData = null;
```
AV29: I don't think it is necessary at all here

src/components/reusable/CityItem/CityItem.jsx
```diff
@@ -0,0 +1,55 @@
...
+
+    constructor(props) {
+        super(props);
+        this.cityData = this.props.cityData;
+        this.handleClick = this.handleClick.bind(this);
+    }
+
+    handleClick() {
+        this.cityData.isCurrentCity = true;
```
AV29: be carefull with mutations... they could become confusing not only for you...

You should just track it via redux. You are already dispatching an action to setCurrent city. Just change respective flag via reducer.

src/components/reusable/CurrentDate/CurrentDate.jsx
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+
+class CurrentDate extends React.Component {
+    today = null;
```
AV29: you don't have to declare these 

you are already assigning them in c-tor

src/components/reusable/CurrentDate/CurrentDate.jsx
```diff
@@ -0,0 +1,25 @@
...
+    date = null;
+
+    constructor() {
+        super();
+        this.today = new Date();
```
AV29: What if you run you App at 11 59: 59 PM ?  within a second new date should be rendered but I doubt that would happen in your case... ;)

src/components/reusable/WeatherIcon/WeatherIcon.jsx
```diff
@@ -0,0 +1,40 @@
...
+class WeatherIcon extends React.Component {
+    weatherIcons = null;
+
+    constructor(props) {
+        super(props);
+        const iconInfo = this.props.iconInfo;
+        const isCurrentCity = this.props.isCurrentCity;
+        this.weather = iconInfo.weather;
+        this.iconType = iconInfo.iconType;
+        this.width = isCurrentCity ? ICON_WIDTH_BIG : ICON_WIDTH_SMALL;
```
AV29: Can you elaborate why you don;t use prop values directly but instead store them on instance in construcotr?

src/reducers/cityWeatherReducer.jsx
```diff
@@ -0,0 +1,67 @@
...
+    allCities: [
+        {
+            weather: "clear-night",
+            iconType: "CLEAR_NIGHT",
+            cityName: "Moscow",
+            temperature: "+5°C",
+            wind: "SE",
+            id: 1,
+        },
+
```
AV29: redundant blank lines ?

src/reducers/cityWeatherReducer.jsx
```diff
@@ -0,0 +1,67 @@
...
+            cityName: "Milan",
+            temperature: "+17°C",
+            wind: "NW",
+            id: 4,
+        } 
+    ]
+};
+
+const cityWeatherReducer = (state = initialState, action) => {
+    console.log("Reducer");
```
AV29: don't forget to remove console.logs ;)

src/containers/MainContainer.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import Main from '../components/blocks/Main/Main';
+
+const MainContainer = (props) => (
```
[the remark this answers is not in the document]
dima-red: fixed

src/components/blocks/Header/Header.jsx
dima-red: It can, but I need a wrapper for better display.

PR #21, opened 2021-03-30
add routing

virzhinia-petkova:


src/common/helpers.js
```diff
@@ -2,3 +2,17 @@ export const kelvinToCelsius = kelvin => {
   const celsius = kelvin - 273;
   return Number(celsius.toFixed(1));
 };
+
+export const groupQueryString = queryString => {
```
AV29, line 6: Could be refactored to:
```
export const groupQueryString = queryString => 
    queryString
        .slice(1, queryString.length)
        .split('&')
        .reduce((acc, query) => {
            const [key, value] = query.split('=');
            acc[key] = value;
            return acc;
          }, {})
```

You don't have to explain each line, especially where code is self explanatory. Better provide a comment at the top, before defining your function with examples of input and ouput of your functon. See how it is done for docs autogenerators such as JSdoc. 


Later we will discuss other means to keep data (an alternative to passing it via query strings). You amy also need this data in other parts of the app, not only on Home page

src/App.jsx
```diff
@@ -14,7 +17,19 @@ const App = () => {
   return (
     <div className="app">
       <Header />
-      {!isLoading ? <Home /> : <p>Getting things ready for you...</p>}
+      {!isLoading ? (
```
AV29, line 20: you may want give the user an ability to contemplate at least something while your data is loading. I mean not only Getting ready message. Meaning, think about moving loading logic inside Home page. Or maybe release Footer and Header so that they could click around while data is laoding.

src/components/Home.jsx
```diff
@@ -1,14 +1,27 @@
...
+  const history = useHistory();
+  const { search } = useLocation();
+
+  // place the query data to key:value pairs
+  const groupedQueryStrings = groupQueryString(search);
+  // replace %20 with spaces
+  const cityName = groupedQueryStrings.city?.replace(/%20/g, ' ');
+  const cityId = groupedQueryStrings.id;
+
+  // if there is a search query - take it as inital value
```
ximet, line 22: Redundant comments

src/components/SearchList.jsx
```diff
@@ -0,0 +1,67 @@
...
+  const handleSelectCity = e => {
+    // find the exact city chosen from the search results list
+    const selectedCityId = e.target.getAttribute('data-target');
+    const selectedCityName = e.target.innerText;
+
+    // update the url with the searched city name & its id
+    history.push(`/home?city=${selectedCityName}&id=${selectedCityId}`);
+  };
+
+  const matchedResults = useMemo(
```
ximet, line 40: for me, better all business logic move to service. @AntonVlasik or @AV29 what do you think?

src/components/SearchList.jsx
```diff
@@ -0,0 +1,67 @@
...
+    () =>
+      // sort by country and then by city
+      [...searchResults].sort(
+        (a, b) => a.country.localeCompare(b.country) || a.name.localeCompare(b.name)
```
ximet, line 44: for me, better all business logic move to service. @AntonVlasik or @AV29 what do you think?

src/components/CustomHooks/useAPIRequest.jsx
```diff
@@ -0,0 +1,32 @@
...
+  if (!typeMap.has(type)) {
+    return [];
+  }
+
+  useEffect(() => {
+    typeMap
+      .get(type)(urlParam)
+      .then(data => setData(data))
+      .catch(error => {
+        areRequestsCanceled(error) || alert(ERRORS.DEFAULT.message);
```
AV29, line 23: if statement would be more readable here: 
```
if(!areRequestsCanceled(error)) {
  alert(ERRORS.DEFAULT.message);
}
```

src/components/Home.jsx
```diff
@@ -30,6 +47,7 @@ const Home = () => {
         return setSearchError('Please provide a search term');
       }
 
+      history.push(`/home/${searchWord}`);
```
AV29, line 50: you could use global state to share searchWord across components instead passing it as a query string. Thus you won't need to parse query later, which is an additional effort. With that said this variant also works.

src/components/Link.jsx
```diff
@@ -1,11 +1,13 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-const Link = ({ place, onClick }) => (
+const Link = ({ place, onClick, id }) => (
   <a
     className="dropdown__list__item"
     onClick={onClick}
```
AV29, line 7: I suggest to just pass all the necessary data to a event handler function and omit data attributes: 
`(event) => onClick({ event, palce, id } )`

src/services/weatherService.js
```diff
@@ -0,0 +1,25 @@
...
+    return Promise.reject(error);
+  }
+};
+
+export const getCurrentWeatherData = async cityId => {
+  try {
+    const { data } = await axios.get(`${BASE_URL}/current/${cityId}`);
+    return data;
+  } catch (error) {
+    return Promise.reject(error);
```
AV29, line 23: Nice! I wonder if you could just `throw error;`...

src/services/weatherService.js
```diff
@@ -0,0 +1,25 @@
...
+  try {
+    const {
+      data: { locations }
+    } = await axios.get(`${BASE_URL}/location/search/${location}`);
+    return [...locations].sort(
+      (a, b) => a.country.localeCompare(b.country) || a.name.localeCompare(b.name)
+    );
+  } catch (error) {
+    // to be handled accordingly elsewhere
+    return Promise.reject(error);
```
AV29, line 14: Great that you added that. This comment will be also a reminder for me to talk about Promises next time.

src/components/SearchList.jsx
```diff
@@ -0,0 +1,38 @@
...
+    const selectedCityId = e.target.dataset.target;
+    const selectedCityName = e.target.dataset.name;
+
+    history.push(`/home?city=${selectedCityName}&id=${selectedCityId}`);
+  };
+
+  return (
+    <div className="app__main">
+      <h2>Here's what we found:</h2>
+      {locations.length > 0 ? (
```
AV29, line 22: At the initial point will the user see something similat to this?
```
<h2>Here's what we found:</h2>
<p>Ooops, looks like we couldn't find anything</p>
```

1) too many words find/found too close to each other.
2) the user will see this text unitl something is actually loaded

If that is the case I'd suggest to work with UX a bit more.

src/components/Home.jsx
```diff
@@ -75,8 +93,10 @@ const Home = () => {
           onClick={handleListItemClick}
         />
       </section>
-      {searchWord ? <CurrentWeather currentSearch={searchWord} /> : null}
+      {search ? <CurrentWeather cityId={cityId} cityName={cityName} /> : null}
     </main>
+  ) : (
+    <p>Getting things ready for you</p>
```
AV29, line 99: I like the wording! Sounds like you really care about your users! That is cool!

src/components/Home.jsx
```diff
@@ -1,16 +1,32 @@
...
+  const cityId = groupedQueryStrings.id;
+
+  const initSearchWordValue = cityName || '';
+  const [searchWord, setSearchWord] = useState(initSearchWordValue);
   const [searchError, setSearchError] = useState('');
   const [storedValue, setValue] = useLocalStorage('searchedPlaces', []);
+  const [isAuthLoading, setIsAuthLoading] = useState(true);
+
+  useEffect(() => {
+    refreshAccessToken().then(() => setIsAuthLoading(false));
```
AV29, line 28: Should you use async await or promises syntax everywhere? Make a dicision :)

src/components/Home.jsx
```diff
@@ -1,16 +1,32 @@
...
+import { groupQueryString } from '../common/helpers';
+import { refreshAccessToken } from '../common/auth';
 
 const Home = () => {
-  const [searchWord, setSearchWord] = useState('');
+  const history = useHistory();
+  const { search } = useLocation();
+
+  const groupedQueryStrings = groupQueryString(search);
+  const cityName = groupedQueryStrings.city?.replace(/%20/g, ' ');
```
AV29, line 18: no magic strings please:) Create a function, say, trim(), or removeSpaces(), or whatever else. Or at least create a constant for your regExp. 

And again... do you really need to use query params to pass around your data ?

src/components/CurrentWeather.jsx
```diff
@@ -1,12 +1,21 @@
...
-  <p>{`The current weather in ${currentSearch} is ${currentTemp} C.`}</p>
-);
+import useAPIRequest from './CustomHooks/useAPIRequest';
+import { REQUEST_TYPES } from '../common/constants';
+
+const CurrentWeather = ({ cityId, cityName }) => {
+  const { current } = useAPIRequest(REQUEST_TYPES.current, cityId);
+
+  return current ? (
+    <p>{`The current weather in ${cityName} is ${current.temperature} C.`}</p>
```
AV29, line 10: Just a hint, maybe overcomplicating but still... you may also handle the situation when for some reason temperature property is not present on current object. What will user see then? 

Plus, the weather may be fine or bad or cloudy, or foggy...but the weather is 10 degrees? I think the wording may be a bit different.

src/common/axios-config.js
```diff
@@ -3,8 +3,16 @@ import { refreshAccessToken } from './auth';
...
-axiosInstance.CancelToken = axios.CancelToken;
+let cancelTokenSource = axios.CancelToken.source();
+
+const cancelRequest = () => {
+  cancelTokenSource.cancel();
+  cancelTokenSource = axios.CancelToken.source();
+};
+
+const areRequestsCanceled = err => {
+  return err && axios.isCancel(err);
```
AV29, line 14: is it possible to omit `err &&` part?  `const areRequestsCanceled = err => axios.isCancel(err)` ?

src/components/Home.jsx
virzhinia-petkova: I'm doing the whole query strings thing so that if the user refreshes the page/sends link to a friend, it will still show the same thing. :) I would've used context or lifted up the state otherwise

src/components/SearchList.jsx
virzhinia-petkova: I fixed that when I connected redux & added isLoading. Is it ok to keep it that way for now?

AV29: sure

src/components/Home.jsx
AV29: Sounds fair!

src/services/weatherService.js
```diff
@@ -11,7 +11,7 @@ export const getLocationData = async location => {
     );
   } catch (error) {
     // to be handled accordingly elsewhere
-    return Promise.reject(error);
+    throw error;
```
AV29, line 14: Your variant was fine though, I didn;t mean to enforce my suggestion :) BTW if you are not handling error in any way here  - then you could jsut omit the catch part and thus your error state will fall through the flow to the next catch itslef. We will talk about it more on our next session.

AV29 (approved):
Good job!!!

PR #22, opened 2021-03-31
Updated branch in local repo with all changes according comments

MartinYDimitrov:


src/components/Button/Button.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styles from './Button.scss';
+
+function Button({ onClick, title }) {
+  return (
+    <button type="button" className={styles.menuButton} onClick={onClick}>
+      {title}
```
AV29, line 8: I'd suggest you to use prop.children approach. That would make your component more flexible. Plus title property usually also needs to be declared separately.

src/components/Header/Header.jsx
```diff
@@ -0,0 +1,74 @@
...
+  }, [isVisible]);
+
+  const changeValue = event => {
+    setInputValue(event.target.value.toLowerCase());
+  };
+
+  return (
+    <header className={styles.header}>
+      <nav className={styles.container}>
+        <Button onClick={handleClick} title="Cities" />
```
AV29, line 62: handleClick might have said a bit more about what this hanlder is doing...

src/components/Header/Header.jsx
```diff
@@ -0,0 +1,74 @@
...
+            setCities(dataCities);
+          });
+      } catch (error) {
+        history.push('/error');
+      }
+    }
+    return () => cancelTokenSource.cancel();
+  }, [inputValue]);
+
+  useEffect(() => {
```
AV29, line 51: Do you really need this effect?

src/components/Header/Header.jsx
```diff
@@ -0,0 +1,74 @@
...
+
+  const changeValue = event => {
+    setInputValue(event.target.value.toLowerCase());
+  };
+
+  return (
+    <header className={styles.header}>
+      <nav className={styles.container}>
+        <Button onClick={handleClick} title="Cities" />
+        {isVisible && (
```
AV29, line 63: could isVisible, setIsVisible pair be removed?

src/components/Header/Header.jsx
```diff
@@ -0,0 +1,74 @@
...
+import { useHistory } from 'react-router';
+
+function Header() {
+  const [isVisible, setIsVisible] = useState(false);
+  const [cities, setCities] = useState([]);
+  const [inputValue, setInputValue] = useState('');
+  const history = useHistory();
+
+  const handleClick = () => {
+    setIsVisible(prev => !prev);
```
AV29, line 17: What is the purpose of this handler?

src/components/Header/Header.jsx
```diff
@@ -0,0 +1,74 @@
...
+  };
+
+  const context = useContext(UserContext);
+  const token = context.token;
+
+  useEffect(() => {
+    const cancelTokenSource = axios.CancelToken.source();
+    if (token !== '' && inputValue !== '') {
```
AV29, line 25: Token related logic is not something that should be put in each request performer. Try to resolve token validity on one centralised place. IF token is expired or about to expire - refresh it.

src/pages/HomePage/HomePage.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import styles from './HomePage.scss';
```
AV29, line 2: Try to come up with a unified order of importing files. Say, Thrid aprty modules come first. Then you components. Then assets. Then styles. 

You can come up with yours, but try to keep it celan and consistent

src/Context/ContextProvider.jsx
```diff
@@ -0,0 +1,35 @@
...
+          console.log(error);
+        });
+    } catch (error) {
+      console.log(error);
+    }
+
+    return () => cancelTokenSource.cancel();
+  }, []);
+
+  return <UserContext.Provider value={{ token }}>{children}</UserContext.Provider>;
```
AV29, line 28: I'm not sure that all children have to know about the token.

src/components/Header/Header.jsx
MartinYDimitrov: Why to remove  isVisible, setIsVisible pair? I think this way is OK.

src/components/Header/Header.jsx
MartinYDimitrov: I will remove Context and ContextProvider(in current task about Redux) and token will be store in Redex Store.

src/Context/ContextProvider.jsx
MartinYDimitrov: As I said i will remove Context and ContextProvider(in current task about Redux).

src/components/Header/Header.jsx
```diff
@@ -0,0 +1,73 @@
...
+  }, [inputValue]);
+
+  const changeValue = event => {
+    setInputValue(event.target.value.toLowerCase());
+  };
+
+  return (
+    <header className={styles.header}>
+      <nav className={styles.container}>
+        <Button onClick={changeFilterableListProperties} children="Cities" />
```
anton-vlasik, line 61: children is something that you usually pass in between opening and closing tags. 
```
 <Button onClick={changeFilterableListProperties} >
    "Cities"
 </Button>   
```


Moreover what exactly would user think when seeing a button, that says "Cities"? Honestly saying I, personally, not feel myself expecting something specific to happen.

src/components/Header/Header.jsx
```diff
@@ -0,0 +1,73 @@
+import React, { useContext, useState, useEffect } from 'react';
+import { useHistory } from 'react-router';
+import axios from 'axios';
+import UserContext from '../../Context/UserContext';
+import FilterableList from '../FilterableList/FilterableList';
+import Button from '../Button/Button';
+import MenuList from '../MenuList/MenuList';
+import styles from './Header.scss';
+
```
anton-vlasik, line 9: Try to keep line gaps consistent too

src/components/Header/Header.jsx
anton-vlasik: The point is that you are changing cities and isVisible thing simultaneously. And basically use then in one place then. Why keep both?

Whould this code produce the same result?

```
{cities.length > 0 && (<FilterableList items={cities} onChange={changeValue} inputValue={inputValue} />)}
```

?

src/components/Header/Header.jsx
```diff
@@ -0,0 +1,73 @@
...
+
+
+function Header() {
+  const [isVisible, setIsVisible] = useState(false);
+  const [cities, setCities] = useState([]);
+  const [inputValue, setInputValue] = useState('');
+  const history = useHistory();
+
+  const changeFilterableListProperties = () => {
```
anton-vlasik, line 18: The name is rather confusing.. .what properties of filterable list do I change? It seems like you are just hiding and revealing your list instead. 

As I'm describing down below... choose one thing to hide your data. It should be enough to either set visible flag or cities. The first one seems more preferrable.

src/components/Button/Button.scss
```diff
@@ -0,0 +1,8 @@
+@import '../../common/variables.scss';
+
+.menuButton {
+  width: 7em;
+  background-color: $tertiary;
```
ximet, line 5: move the color to variables!

src/components/Header/Header.jsx
```diff
@@ -0,0 +1,73 @@
...
+  };
+
+  const context = useContext(UserContext);
+  const token = context.token;
+
+  useEffect(() => {
+    const cancelTokenSource = axios.CancelToken.source();
+    if (token !== '' && inputValue !== '') {
+      try {
+        axios
```
ximet, line 30: move business logic to service

src/pages/HomePage/HomePage.jsx
```diff
@@ -1,12 +1,34 @@
...
 
 function HomePage() {
+  const [cities, setCities] = useState([]);
+  const [inputValue, setInputValue] = useState('');
+  const history = useHistory();
+
+  useEffect(async () => {
+    if (inputValue.length >= 3) {
+      try {
+        const getDataCites = await getCitiesData(inputValue);
```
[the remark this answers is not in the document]
anton-vlasik: Could this code be moved to changeValue handler? do you need an effect?

src/components/FilterableList/FilterableList.jsx
```diff
@@ -6,7 +6,12 @@ function FilterableList({ items, onChange, inputValue }) {
   return (
     <ul className={styles.list}>
       <li className={styles.listItem}>
-        <input type="text" className={styles.searchBox} onChange={onChange} />
+        <input
+          type="text"
+          className={styles.searchBox}
+          onChange={onChange}
+          placeholder={'Enter city name'}
```
anton-vlasik, line 13: no need to use curly braces in this case

src/common/getToken.js
```diff
@@ -0,0 +1,12 @@
+import axios from 'axios';
+
+export async function getToken() {
+  try {
+    const res = await axios.get('http://localhost:3000/auth');
```
anton-vlasik, line 5: magic string!

src/pages/HomePage/HomePage.jsx
```diff
@@ -1,12 +1,34 @@
...
+import FilterableList from '../../components/FilterableList/FilterableList';
+import getCitiesData from '../../services/getCitiesData';
 
 function HomePage() {
+  const [cities, setCities] = useState([]);
+  const [inputValue, setInputValue] = useState('');
+  const history = useHistory();
+
+  useEffect(async () => {
+    if (inputValue.length >= 3) {
```
anton-vlasik, line 13: Magic number!!!

src/common/getToken.js
ximet: I don't understand what this getToken method. Wrapper for wrapper?

src/components/FilterableList/FilterableList.jsx
ximet: for me, all string constant need to move to some const.

src/components/FilterableList/FilterableList.scss
```diff
@@ -0,0 +1,29 @@
+@import '../../common/variables.scss';
+
+.list {
+  margin: 0;
+  list-style-type: none;
+  position: fixed;
+  top: $headerHeight;
+  left: 0;
+  padding: 0 10px;
```
ximet, line 9: again PX.... but you're using em

src/components/Footer/Footer.scss
```diff
@@ -0,0 +1,13 @@
...
+.footer {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  background-color: $secondary;
+  height: $footerHeight;
+  position: fixed;
+  bottom: 0;
+  left: 0;
+  width: 100%;
```
ximet, line 12: em or %?

src/common/getToken.js
MartinYDimitrov: To get token from my node server and set it in local storage.

src/components/Footer/Footer.scss
MartinYDimitrov: My idea is the footer to be responsive to the window and always to be 100% of the window width. If I use em the width of footer will be static. If the idea is to be consistent I prefer to change the value of height property to %.

src/pages/HomePage/HomePage.jsx
```diff
@@ -0,0 +1,36 @@
...
+  const [cities, setCities] = useState([]);
+  const [inputValue, setInputValue] = useState('');
+  const history = useHistory();
+
+  const getAllCitiesByName = async event => {
+    setInputValue(event.target.value.toLowerCase());
+
+    if (inputValue.length >= minValueLength) {
+      try {
+        let getDataCities = await getCitiesData(inputValue);
```
anton-vlasik, line 18: 1) why let instead of const? Are you going to mutate it later?
2) getDataCities - is a name for a function - it is represented by a verb: get something. Whereas in this case you should use a noun: const cities... or const citiesData

src/pages/HomePage/HomePage.jsx
```diff
@@ -0,0 +1,36 @@
...
+        setCities(getDataCities);
+      } catch (error) {
+        console.log(error);
+        history.push('/error');
+      }
+    } else {
+      setCities([]);
```
anton-vlasik, line 25: should you check the fact that cities are already set to [] to reduce number of updates ?

MartinYDimitrov: If the user clear the input value I want cities to be set it to [] and FilterableList component to be re-rendered with empty array not to be shown with previous state.

src/pages/HomePage/HomePage.jsx
```diff
@@ -10,25 +10,28 @@ function HomePage() {
   const [inputValue, setInputValue] = useState('');
   const history = useHistory();
 
-  const getAllCitiesByName = async event => {
-    setInputValue(event.target.value.toLowerCase());
-
+  useEffect(async () => {
```
MartinYDimitrov, line 13: In previous comments you ask me is I really need the effect. Actually I need this effect, because If I made the request to API from changeValueHandler (getAllCitiesByName) a bug appears. The request to API is made with previous value of inputValue. For example if user write 'Sofia' the request is made with 'Sofi'.

src/pages/HomePage/HomePage.jsx
```diff
@@ -0,0 +1,35 @@
...
+import PageLayout from '../../PageLayout/PageLayout';
+import FilterableList from '../../components/FilterableList/FilterableList';
+import getCitiesData from '../../services/getCitiesData';
+import { minValueLength, zeroLength } from '../../common/constants';
+
+function HomePage() {
+  const [cities, setCities] = useState([]);
+  const history = useHistory();
+
+  const getAllCitiesByName = async event => {
```
anton-vlasik, line 12:
```
 const getAllCitiesByName = async event => {
    const query= event.target.value.toLowerCase();

    if (query.length >= minRequestingQueryLength) {
      try {
        const citiesData= await getCitiesData(query);
        setCities(citiesData);
      } catch (error) {
        console.error(error);
        history.push('/error');
      }
    }
    
    if (query.length === 0 && cities.length > 0) {
       setCities([]);
    }
}
```

src/Context/ContextProvider.jsx
```diff
@@ -0,0 +1,35 @@
...
+import UserContext from './UserContext';
+
+function ContextProvider({ children }) {
+  const [token, setToken] = useState('');
+
+  useEffect(async () => {
+    const cancelTokenSource = axios.CancelToken.source();
+
+    try {
+      const res = await axios
```
ximet, line 13: move businessLogic to service

AV29 (approved):

PR #26, opened 2021-04-14
Adding selectors

virzhinia-petkova:


MartinYDimitrov (approved):
I don't have remarks!

package.json
```diff
@@ -29,6 +29,7 @@
     "babel-plugin-transform-react-remove-prop-types": "0.4.24",
     "core-js": "3.2.1",
     "history": "4.10.1",
+    "lodash.isequal": "^4.5.0",
```
ximet, line 32: for what

ximet (approved):

src/redux/selectors/locationsSelectors.js
```diff
@@ -0,0 +1,15 @@
...
+import isEqual from 'lodash.isequal';
+
+const locationsSelector = state => state.locations.locations;
+export const isLocationsLoadingSelector = state => state.locations.isLoading;
+
+// I realize this is a bit of overkill, just wanted to try out the selectors
+const createDeepEqualSelector = createSelectorCreator(defaultMemoize, isEqual);
+
+export const transformLocations = createDeepEqualSelector(locationsSelector, locations =>
+  locations?.reduce((acc, location) => {
```
anton-vlasik, line 11: correct me if I'm wrong: are locations taken from the initial state of redux? if so  - then they should be defined as [] in a worst case scenario. 

also... is this a much better situation for map, rather than reduce? 

```
locations.map(location => ({...location, fullName: `${location.name}, ${location.country}` });
```

ximet, line 11: you can use .map

virzhinia-petkova: They're null in the initial state, but I can fix that

package.json
virzhinia-petkova: I use it to create a deep equality selector, I could write my own function to check if they're equal

src/redux/selectors/locationsSelectors.js
```diff
@@ -0,0 +1,12 @@
+import { createSelectorCreator, defaultMemoize } from 'reselect';
+import { areLocationsTheSame } from '../../common/helpers';
+
+const locationsSelector = state => state.locations.locations;
+export const isLocationsLoadingSelector = state => state.locations.isLoading;
+
+// I realize this is a bit of overkill, just wanted to try out the selectors
+const createDeepEqualSelector = createSelectorCreator(defaultMemoize, areLocationsTheSame);
```
AV29, line 8: so you can't apply your selector for other means within locations feature anymore. You are creating a super specific selector fabric that can only work based on areLocationsTheSame comparator. I don't think this is a good idea.

src/components/CurrentWeather/CurrentWeatherContainer.js
```diff
@@ -1,14 +1,17 @@
 import { connect } from 'react-redux';
 
 import CurrentWeather from './CurrentWeather';
+
```
AV29, line 4: opinionated, but feeling super iffy about theese gaps :)

ximet: agree

src/redux/selectors/locationsSelectors.js
ximet: agree

anton-vlasik (approved):
keep in mind suggested approach for naming selectors: 
1) getters from the state are  - getSomething
2) actual selectors are - either selectSomething, or even (I prefer this more) deriveSomething

Anyway make sure it is consistent throughout the whole project. So that other devs could totally distinguish one thing from another and from simple helper functions

PR #27, opened 2021-04-16
reduxStore

MartinYDimitrov:


src/App.jsx
```diff
@@ -1,8 +1,14 @@
 import React from 'react';
+import { Provider } from 'react-redux';
 import Navigation from './Navigation/Navigation';
+import store from './ReduxStore/store/index';
```
AV29, line 4: don't have to specify index here

src/ReduxStore/reducers/currentCity.js
```diff
@@ -0,0 +1,50 @@
...
+      return {
+        ...state,
+        isLoaded: true,
+        city: action.payload
+      };
+    }
+    case CURRENT_CITY_FAIL: {
+      return {
+        ...state,
+        isLoaded: false,
```
AV29, line 33: a little bit confusing for me. It is actually loaded. The request is finished. Failure may also be a valid response, say, 404

src/services/dataServices.js
```diff
@@ -0,0 +1,9 @@
+import getCitiesData from './getCitiesData';
+import getData from './getData';
+
+const dataServices = {
```
AV29, line 4: I'd say it is the one service containing two methods. getCitiesData is more like a method for the service

src/common/constants.js
```diff
@@ -1,6 +1,7 @@
 export const authURL = 'http://localhost:3000/auth';
 export const baseURL = 'https://pfa.foreca.com/api/v1';
 export const locationURL = '/location/search/';
+export const currentURL = '/current/';
```
AV29, line 4: trailing slash seems redundant

src/ReduxStore/actions/currentCity.js
```diff
@@ -0,0 +1,20 @@
...
+export const CURRENT_CITY_LOADING = `${PREFIX}CURRENT_CITY_LOADING`;
+export const CURRENT_CITY_LOADED = `${PREFIX}CURRENT_CITY_LOADED`;
+export const CURRENT_CITY_FAIL = `${PREFIX}CURRENT_CITY_FAIL`;
+export const SET_CURRENT_CITY_NAME = `${PREFIX}SET_CURRENT_CITY_NAME`;
+
+export function getCurrentCityData(id) {
+  return async function (dispatch, getState, dataServices) {
+    try {
+      dispatch({ type: CURRENT_CITY_LOADING });
+      const data = await dataServices.getData(`${currentURL}${id}`);
```
virzhinia-petkova, line 13: in the getData method you're passing a slice of the url here while in the getCitiesData you're only passing the location and handling the rest of the url in the service. I'd stick to one of those approaches so that it's more consistent how the methods are supposed to be used

AV29: Good call. I think is it no place here to use urls - move them to service at least. The onyl thing you better do here is 
`await dataServices.getData(id or params)`

virzhinia-petkova (approved):

src/ReduxStore/reducers/currentCity.js
```diff
@@ -0,0 +1,50 @@
...
+      };
+    }
+    case CURRENT_CITY_LOADED: {
+      return {
+        ...state,
+        isLoaded: true,
+        city: action.payload
+      };
+    }
+    case CURRENT_CITY_FAIL: {
```
AV29, line 30: Honestly saying I don't like this name. It is not your city that failed, but the loading of it's data. So it is better to name this action `CURRENT_CITY_LOADING_FAILED`.... or similar.

src/ReduxStore/reducers/matchedCities.js
```diff
@@ -0,0 +1,48 @@
...
+      };
+    }
+    case MATCHED_CITIES_DATA_LOADED: {
+      return {
+        ...state,
+        isLoaded: true,
+        cities: action.payload
+      };
+    }
+    case MATCHED_CITIES_DATA_FAIL: {
```
AV29, line 29: Same here

AV29 (approved):

PR #29, opened 2021-04-19
Task/selectors

MartinYDimitrov:
Added selectors

### DIFF
--- a/src/ReduxStore/selectors/matchedCitiesSelectors.js
+++ b/src/ReduxStore/selectors/matchedCitiesSelectors.js
@@ -2,12 +2,12 @@ import { createSelector } from 'reselect';
 
 export const getMatchedCitiesData = state => state.matchedCities.cities;
 
-export const transformCitiesData = createSelector(getMatchedCitiesData, cities => {
-  return [...cities]
-    .sort((a, b) => a.country.localeCompare(b.country) || a.name.localeCompare(b.name))
+export const selectMatchedCitiesData = createSelector(getMatchedCitiesData, cities => {
+  return cities
     .map(city => ({
       name: `${city.name.toLowerCase()}`,
       country: `${city.country.toLowerCase()}`,
       id: city.id
-    }));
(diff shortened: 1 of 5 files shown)


src/ReduxStore/selectors/matchedCitiesSelectors.js
```diff
@@ -0,0 +1,13 @@
+import { createSelector } from 'reselect';
+
+export const getMatchedCitiesData = state => state.matchedCities.cities;
+
+export const transformCitiesData = createSelector(getMatchedCitiesData, cities => {
+  return [...cities]
```
anton-vlasik, line 6: you can do the mapping first so that you won't have to additionally create new array

src/ReduxStore/selectors/matchedCitiesSelectors.js
```diff
@@ -0,0 +1,13 @@
+import { createSelector } from 'reselect';
+
+export const getMatchedCitiesData = state => state.matchedCities.cities;
+
+export const transformCitiesData = createSelector(getMatchedCitiesData, cities => {
```
virzhinia-petkova, line 5: I believe this one is supposed to be selectCitiesData

anton-vlasik, line 5: The only thing here I'd suggest to prefix selectors (made with createSelector utility) with some unified thing, like: selectSomething, or deriveSomething

src/services/getCitiesData.js
```diff
@@ -4,11 +4,7 @@ import getData from './getData';
 export default async function getCitiesData(query) {
   try {
     const { locations } = await getData(`${locationURL}${query}`);
```
anton-vlasik, line 6: It is still fine to have a layer, kinda DTO where you prepare the data after fetch. But just make sure that you are not filtering or corrupting useful data on this stage.

anton-vlasik (approved):

PR #30, opened 2021-04-22
Adding themes

virzhinia-petkova:


src/components/App/style.js
```diff
@@ -0,0 +1,30 @@
...
+
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    justify-content: space-between;
+
+    font-size: ${theme.fontSize.default};
+    color: ${theme.colors.fontColor};
+
+    &__footer {
```
MartinYDimitrov, line 19: I think you forgot to remove those. I think it is scss.

package.json
```diff
@@ -38,7 +38,8 @@
     "redux": "4.0.5",
     "redux-thunk": "2.3.0",
     "reselect": "4.0.0",
-    "scss": "^0.2.4"
+    "scss": "^0.2.4",
```
ximet, line 41: "It's a training, so I'd say let's stick to one approach" (© Anton Vlasik)

src/components/Error.jsx
```diff
@@ -1,15 +1,14 @@
 import { useHistory } from 'react-router-dom';
 import errorPhoto from '../../public/404.jpg';
+import * as S from './Theme/GlobalStyles';
```
ximet, line 3: I don't like naming like S. It's not obvious

package.json
virzhinia-petkova: I'm not using it, I just haven't uninstalled it. I'll do that

src/components/Header/style.js
```diff
@@ -0,0 +1,17 @@
...
+
+const styledHeader = styled(Header)`
+  ${({ theme }) => `
+  ${console.log(theme)}
+        flex: 1;
+        padding: ${theme.padding.default} 0;
+        width: inherit;
+        max-height: ${theme.height.s};
+    
+        display: flex;
```
AV29, line 13: indentation of the whole block looks weird

src/components/Home/Home.jsx
```diff
@@ -0,0 +1,106 @@
...
+import CurrentWeather from '../CurrentWeather/CurrentWeatherContainer';
+
+import useLocalStorage from '../CustomHooks/useLocalStorage';
+import { ENTER_KEYCODE, MAX_ITEMS_LENGTH } from '../../common/constants';
+import { groupQueryString, transformSpaces } from '../../common/helpers';
+import { refreshAccessToken } from '../../common/auth';
+import * as S from '../Theme/GlobalStyles';
+
+const Home = ({ className }) => {
+  const history = useHistory();
```
AV29, line 15: a lot of stuff going on here. Hard to distinguish. I suggest to start creating abstactions like custom hooks here

src/components/Theme/GlobalStyles.js
```diff
@@ -0,0 +1,169 @@
...
+            outline: none;
+        }`}
+`;
+
+export const Navigation = styled.nav`
+  list-style: none;
+  display: flex;
+`;
+
+export const Links = styled(Link)`
```
AV29, line 78: singular?

src/components/Theme/Theme.jsx
```diff
@@ -0,0 +1,80 @@
...
+    default: '0.75em'
+  },
+  backgroundImage: {
+    image: `url(${searchIcon})`,
+    size: '1em',
+    position: '0.75em'
+  }
+});
+
+const Root = () => {
```
AV29, line 67: The file is named Theme, the export thing is Root... looks like you need to reorganize things here

src/components/Theme/Theme.jsx
```diff
@@ -0,0 +1,80 @@
...
+  const [storedValue, setValue] = useLocalStorage('mode', THEMES.light);
+  const handleToggle = () => setValue(storedValue === THEMES.light ? THEMES.dark : THEMES.light);
+  const mode = themesMap.get(storedValue);
+
+  return (
+    <ThemeProvider theme={getTheme(mode)}>
```
AV29, line 73: I'd leave Root as is as well as the name of the module but move all Theme related components to a separate folder

AV29 (commented):
I noticed that some components are declared as StyledSomething, and some of those are exported as is and then imported with help of * as S. Is there any pattern there that you intended to implement?

src/components/Error.jsx
anton-vlasik: it may or may not be a convention on a project. But could be easily changed to Styled, right @virzhinia-petkova ? :)

virzhinia-petkova: I already changed it to exactly Styled :)

virzhinia-petkova:
> I noticed that some components are declared as StyledSomething, and some of those are exported as is and then imported with help of * as S. Is there any pattern there that you intended to implement?

I did generic elements like buttons and inputs with * as S, while entire views (Home, About, etc) are used like styled(Component) with StyledComponent so that I don't have name collisions with the `Component`  that has been imported, then I'm passing the generated className as props. It made sense for me to separate them. Would it be better to only use one of those? Maybe use everything with * as S and not have the styled(Component) type of thing?

src/styles/theme.js
```diff
@@ -0,0 +1,61 @@
...
+};
+
+export const themesMap = new Map([
+  [THEMES.dark, dark],
+  [THEMES.light, light]
+]);
+
+export const getTheme = mode => ({
+  colors: mode,
+  fontSize: {
```
AV29, line 31: not sure about keeping all the static values within theme object. You are able to change your small width value in one place and it will be changed everywhere.... it is for you to decide whether it is super good or super bad. Do not overconfigure it.

virzhinia-petkova: Keeping all values in variables was an idea we discussed with Igor, when he was our trainer, that I really liked.  The biggest benefit for me is the fact that I don't need to remember/look up values in order to create consistent styles

AV29: ok, if it makes your work easier - that's good. But I really doubt that you never look up there when choosing between, say,  s and xs.

AV29 (approved):

PR #42, opened 2021-09-21
created a server with the ability to obtain an access token from weather Api

DmitriyKrivovyaz:
…her api


server/package.json
```diff
@@ -0,0 +1,19 @@
...
+    "dev": "nodemon index.js"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "axios": "^0.21.4",
+    "cors": "^2.8.5",
+    "dotenv": "^10.0.0",
+    "express": "^4.17.1",
+    "nodemon": "^2.0.12",
```
glebdronchenko, line 16: This package need just for devDependencies.

glebdronchenko:
Why do you create axios connection in index file from server/src/http/ and then you import this file in weatherApi file from the same folder? Maybe you need create weatherApiService file and move this all functionality there.

What does your comment in this PR mean?

glebdronchenko (commented):
Why do you create axios connection in index file from server/src/http/ and then you import this file in weatherApi file from the same folder? Maybe you need create weatherApiService file and move this all functionality there.

What does your comment in this PR mean?

PR #46, opened 2021-09-22
created server to obtain an access token from weather api

DmitriyKrivovyaz:


server/index.js
```diff
@@ -0,0 +1,27 @@
...
+
+const app = express();
+
+app.use(cors())
+app.use(express.json())
+app.use('/api', router)
+
+app.get( '/', (req, res) => {
+    res.status(200).json({message: 'WORKING'})
+})
```
glebdronchenko, line 16: What is this route for here?

server/src/http/index.js
```diff
@@ -0,0 +1,10 @@
+const axios = require('axios');
+require('dotenv').config();
+
+const WEATHER_API_URL = process.env.WEATHER_API_URL;
+
+const $host = axios.create({
```
glebdronchenko, line 6: Why do you use $ symbol? What does it mean?

glebdronchenko (commented):
Why do you create axios connection in index file from server/src/http/ and then you import this file in weatherApi file from the same folder? Maybe you need create weatherApiService file and move this all functionality there.

glebdronchenko (approved):

PR #52, opened 2021-10-04
Login page

usermeme:


src/app/layouts/public-layout/PublicLayout.jsx
```diff
@@ -1,7 +1,12 @@
-import { Suspense } from 'react';
+import * as React from 'react';
+import * as Components from 'components';
 
 const PublicLayout = ({ children }) => (
-  <Suspense fallback={null}>{children}</Suspense>
+  <Components.Wrapper>
```
anton-vlasik: TBH, it's a bit hard to say what's a wrapper and what's a container

src/app/providers/app-provider/AppProvider.jsx
```diff
@@ -1,8 +1,18 @@
...
   const isLoggedIn = useSelector((state) => state.auth.isLoggedIn);
 
+  const dispatch = useDispatch();
+
+  React.useEffect(() => {
+    const token = localStorage.getItem(TOKEN);
+    if (token) dispatch(setLoggedIn(true));
+  }, [dispatch]);
+
   return <>{children(isLoggedIn)}</>;
```
anton-vlasik: not sure if you need Fragment at all. You could just return children(isLoggedIn)

Also alternatively (if you choose to go with hooks) you could create custom hook...pseudo code:

```
const useAuth = () => {
  const isLoggedIn = useSelector((state) => state.auth.isLoggedIn);

  const dispatch = useDispatch();

  React.useEffect(() => {
    const token = localStorage.getItem(TOKEN);
    if (token) dispatch(setLoggedIn(true));
  }, [dispatch]);
  
  return isLoggedIn;
```

Later you ma have used it like that: 

```
const App = () => {
   const isLoggedIn = useAuth();
   return isLoggedIn ? <...> : <...>
};
```

at the end of the day I personally think that your FACC component is not so AppProvider, though the name itself is pretty vague. It seems like it is more of a AuthProvider

src/pages/public/login/components/form/Form.jsx
```diff
@@ -0,0 +1,49 @@
...
+          {t('login_title')}
+        </Components.Button>
+      </form>
+    );
+  }
+}
+
+const mapStateToProps = (state) => ({ loading: state.auth.loading });
+
+const ReduxForm = reduxForm({ form: 'login' })(Form);
```
anton-vlasik, line 46: Is it restricting wrapping functional components ?

src/services/http.js
```diff
@@ -0,0 +1,11 @@
+import axios from 'axios';
+
+const http = axios.create();
+
+http.defaults.baseURL = process.env.BASE_URL;
+http.defaults.headers = {
+  // TODO: change to correct
+  'Access-Control-Allow-Origin': '*',
```
anton-vlasik: This is supposed to be handled on the responding side not from the requesting party. Do I miss something?

anton-vlasik (approved):

PR #53, opened 2021-10-04
Data service

alvikalvik:


src/services/dataService.js
```diff
@@ -0,0 +1,67 @@
...
+      this.token = response.data.access_token;
+    } catch (error) {
+      console.log(error);
+      this.token = null;
+    }
+
+    return this.token;
+  },
+
+  async getData(dataType, param) {
```
anton-vlasik: what if there are more params?

src/services/dataService.js
```diff
@@ -0,0 +1,67 @@
...
+  getAuthHeaders() {
+    return {
+      headers: {
+        Authorization: `Bearer ${this.token}`
+      }
+    };
+  },
+  async getToken(user, password) {
+    const body = {
+      user: `${user}`,
```
anton-vlasik: the same as 
```
const body = {
  user: user, 
  password: password 
}
```
?

src/constants/constants.js
```diff
@@ -3,3 +3,34 @@ export const THEME_DARK = 'dark';
 
 export const CONTACT_US_PAGE_LINK = '/about';
 export const CITIES_PAGE_LINK = '/cities';
+
+export const API_AUTH_USERNAME = 'coherentsolutions';
+export const API_AUTH_PASS = 'X4AbD4NrEd30';
```
anton-vlasik: This is not a good idea! :) I encourage you to read about .env фтв вщеутм package in particualr

src/services/dataService.js
alvikalvik: getData is general function, it fetches different variants of data form url with param. If exact getting data function (such as getForecast) needs some more params it concats all the params into one string param and call getData with this concated param.
If it's bad idea please advise how to do this in a better way.

src/services/dataService.js
alvikalvik: true, the same :)
fixed

src/constants/constants.js
alvikalvik: dotenv-webpack and .env file added
thanks

src/services/dataService.js
anton-vlasik: To my mind this is an overcomplication and overengeneering: I'd say to just have separate functions like that: 
```
async searchLocation(query) {
    ...
    try {
      const response = await instance.get(...here you will pass url, query and other stuff within axios config);
      return response.data[dataKey];
    } catch (error) {
      console.error(error);
    }
},

etc
```

alvikalvik: thanks, changed

src/services/dataService.js
```diff
@@ -0,0 +1,88 @@
...
+    } catch (error) {
+      console.log(error);
+      return null;
+    }
+  },
+
+  async getCurrentWeather(locationId) {
+    const { url, dataKey } = DATA_TYPES[API_GET_CURRENT_WEATHER_DATA_TYPE];
+    try {
+      const response = await instance.get(url + locationId, this.getAuthHeaders());
```
anton-vlasik: why concatenate, could you use params object instead?
```
axios.get('/user', {
    params: {
      ID: 12345
    }
})
```

taken from axios docs

alvikalvik: did not know about this syntax, thanks. But in this case it's not suitable because I need for example to get '/url/locationId' instead of '/url?locationId=locationIdValue'.
So I think it's better to use manually concatenation because it gives flexibility.

ximet (approved):

src/services/dataService.js
anton-vlasik: Ok, got it. Could you use ES6 strings instead? It makes it a bit more readable and more consistent with the rest parts of the project

src/services/dataService.js
```diff
@@ -0,0 +1,80 @@
...
+} from '../constants/constants';
+
+const instance = axios.create({
+  withCredentials: true,
+  baseURL: API_BASE_URL
+});
+
+export const weatherAPI = {
+  token: null,
+  getAuthHeaders() {
```
anton-vlasik: I would rather name this function getHeaders. There may be also other headers later. Plus, why should it return them wrapped in another object? (see comment below)

src/services/dataService.js
```diff
@@ -0,0 +1,80 @@
...
+      this.token = null;
+    }
+
+    return this.token;
+  },
+
+  async searchLocation(query) {
+    const { url, dataKey } = DATA_TYPES[API_SEARCH_LOCATION_DATA_TYPE];
+    try {
+      const response = await instance.get(`${url}${query}`, this.getAuthHeaders());
```
anton-vlasik: Here the second parameter is a bit misleading. It is NOT "headers". It is config that holds headers among other stuff. So I would refactor:

```
getHeaders() {
  return {
        Authorization: `Bearer ${this.token}`
    };
}
```
and 

```
await (http or axios).get(`${url}${locationId}`, { headers : this.getHeaders() });
```

alvikalvik: changed

anton-vlasik (approved):

PR #54, opened 2021-10-04
Basic structure

JaneEllison:
Recreate pull request without .env file and with structure corrected according to comments from [pull request#51](https://github.com/ximet/ReactTask/pull/51)


src/components/NavLink/NavLink.jsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+function NavLink() {
```
alvikalvik: I think it's better to change NavLink name to something else because react router dom library has component with the same name

src/layouts/FavoriteSities/FavoriteSities.jsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+function FavoriteSities() {
```
alvikalvik: Cities maybe?

JaneEllison: Oh. Yes. Fixed

AV29 (approved):

PR #55, opened 2021-10-05
Location weather component

alvikalvik:
I Temporary made data fetching inside LocationWeather component. In future with redux applying I'll move it.


src/constants/constants.js
```diff
@@ -3,3 +3,57 @@ export const THEME_DARK = 'dark';
...
+
+export const API_AUTH_USERNAME = process.env.API_AUTH_USERNAME;
+export const API_AUTH_PASS = process.env.API_AUTH_PASS;
+export const API_BASE_URL = '/api'; //proxed to 'https://pfa.foreca.com' via webpack
+export const API_AUTH_URL = '/authorize/token';
+export const API_SEARCH_URL = '/api/v1/location/search/';
+export const API_GET_LOCATION_INFO_URL = '/api/v1/location/';
+export const API_CURRENT_WEATHER_URL = '/api/v1/current/';
+export const API_FORECAST_URL = '/api/v1/forecast/';
+export const API_FORECAST_3_HOURLY = '3hourly/';
```
anton-vlasik: I have a gut feeling that you can pass any number here to get that much hourly forecast - is that so? Coz if so I'd parametrize this instead.

src/components/LocationWeather/LocationWeather.jsx
```diff
@@ -0,0 +1,85 @@
...
+        <LocationWeatherCurrent
+          currentLocationWeather={this.state.currentLocationWeather}
+          currentLocationInfo={this.state.currentLocationInfo}
+        />
+        <LocationWeatherDailyList
+          currentLocationDailyWeather={this.state.currentLocationDailyWeather}
+          activeDayDate={this.state.activeDayDate}
+          setActiveDayDate={this.setActiveDayDate}
+        />
+        <LocationWeather3HourlyList
```
anton-vlasik: I feel iffy about these 3 hours thing. Could it be parametrized instead? Could you have
```
<LocationWeatherHourlyList length={3} />
```
something like that

Or in the worst case scenario just omit 3 in the name. It would still make it hourlyforecast. It is just your app displays only 3 hours.


That applies to the whole PR

alvikalvik: I renamed all the "3_hourly" things to "detailed". I refused using "hourly" because the interval can be less than a hour. In case of change needed interval I'll have to change only 2 things - API_FORECAST_DETAILED endpoint string and dataFilter function inside LocationWeatherDetailedList component.

src/components/LocationWeather/LocationWeatherCurrent/LocationWeatherCurrent.jsx
```diff
@@ -0,0 +1,66 @@
...
+  API_SYMBOL_URL_PREFIX,
+  API_WIND_IMG_URL_POSTFIX,
+  API_WIND_IMG_URL_PREFIX,
+  WIND_DIRECTIONS_TO_IMG
+} from '../../../constants/constants';
+import { formatTemperature } from '../../../utils/utils';
+import './LocationWeatherCurrent.css';
+
+function LocationWeatherCurrent({ currentLocationWeather, currentLocationInfo }) {
+  const locationName = currentLocationInfo ? currentLocationInfo.name : null;
```
anton-vlasik: why null but not empty string?

src/components/LocationWeather/LocationWeatherDailyList/LocationWeatherDailyList.jsx
```diff
@@ -0,0 +1,31 @@
...
+      <LocationWeatherDailyListItem
+        key={dayData.date}
+        dayData={dayData}
+        activeDayDate={activeDayDate}
+        setActiveDayDate={setActiveDayDate}
+      />
+    );
+  }
+
+  const dailyList = currentLocationDailyWeather
```
anton-vlasik: too many ternaries with the same condition... Isn't that the same?  (also it is a common thing to put mappings inline)
```
return (
    <ul className='location-weather__daily-list'>
      {currentLocationDailyWeather
        ? currentLocationDailyWeather.map(dayData => (
            <LocationWeatherDailyListItem
              key={dayData.date}
              dayData={dayData}
              activeDayDate={activeDayDate}
              setActiveDayDate={setActiveDayDate}
            />
          ))
        : null}
    </ul>
  );
```

src/components/LocationWeather/LocationWeatherDailyListItem/LocationWeatherDailyListItem.jsx
```diff
@@ -0,0 +1,39 @@
...
+function LocationWeatherDailyListItem({ dayData, activeDayDate, setActiveDayDate }) {
+  const isActive = activeDayDate === dayData.date;
+  const date = new Date(dayData.date);
+  const dayName = date.toLocaleString('en-us', { weekday: 'long' });
+  const dayNumber = date.getDate();
+  const month = date.toLocaleString('default', { month: 'long' });
+  return (
+    <li
+      className={`location-weather__daily-list-item${
+        isActive ? ' location-weather__daily-list-item_active' : ''
```
anton-vlasik: hmm...still don't want to use .join

src/components/LocationWeather/LocationWeatherDailyListItem/LocationWeatherDailyListItem.jsx
```diff
@@ -0,0 +1,39 @@
+import { API_SYMBOL_URL_POSTFIX, API_SYMBOL_URL_PREFIX } from '../../../constants/constants';
+import { formatTemperature } from '../../../utils/utils';
+import './LocationWeatherDailyListItem.css';
+
+function LocationWeatherDailyListItem({ dayData, activeDayDate, setActiveDayDate }) {
+  const isActive = activeDayDate === dayData.date;
```
anton-vlasik: you only use activeDayDate to determine if it is active. Would it be better to just pass a property form above, say, isActive={activeDayDate === dayData.date} ?

src/components/LocationWeather/LocationWeatherDailyListItem/LocationWeatherDailyListItem.jsx
```diff
@@ -0,0 +1,39 @@
...
+      <div className="location-weather__daily-list-item-day-number">{dayNumber}</div>
+      <div className="location-weather__daily-list-item-month">{month}</div>
+      <img
+        src={`${API_SYMBOL_URL_PREFIX}${dayData.symbol}${API_SYMBOL_URL_POSTFIX}`}
+        alt={dayData.date}
+        className="location-weather__daily-list-item-symbol-img"
+      />
+      <div className="location-weather__daily-list-item-temperature">
+        <div className="location-weather__daily-list-item-minmax-temperature">
+          {formatTemperature(dayData.minTemp) + '°'}
```
anton-vlasik: try to stick to ES6 strings

src/components/LocationWeather/LocationWeatherDetailedList/LocationWeatherDetailedList.jsx
```diff
@@ -0,0 +1,25 @@
+import LocationWeatherDetailedListItem from '../LocationWeatherDetailedListItem/LocationWeatherDetailedListItem';
+
+import './LocationWeatherDetailedList.css';
+
+function LocationWeatherDetailedList({ currentLocationDetailedWeather, activeDayDate }) {
+  function mapData(detailedData) {
+    return <LocationWeatherDetailedListItem key={detailedData.time} detailedData={detailedData} />;
+  }
+
+  function filterData(detailedData) {
```
anton-vlasik: this could be done inline. It would still remain readable. And it is not filterData per se. It is a predicate function that returns you boolean value. It's not returning you anything filtered.

src/components/LocationWeather/LocationWeatherDetailedListItem/LocationWeatherDetailedListItem.jsx
```diff
@@ -0,0 +1,41 @@
...
+  API_SYMBOL_URL_PREFIX,
+  API_WIND_IMG_URL_POSTFIX,
+  API_WIND_IMG_URL_PREFIX,
+  WIND_DIRECTIONS_TO_IMG
+} from '../../../constants/constants';
+import { formatTemperature } from '../../../utils/utils';
+import './LocationWeatherDetailedListItem.css';
+
+function LocationWeatherDetailedListItem({ detailedData }) {
+  const time = detailedData.time.split('T')[1].split(/[+-]/)[0];
```
anton-vlasik: That looks very ad-hoc-ish. Are you sure you can really rely on that sort of input data being formatted?

src/components/LocationWeather/LocationWeatherDetailedListItem/LocationWeatherDetailedListItem.jsx
```diff
@@ -0,0 +1,41 @@
...
+  return (
+    <li className="location-weather__detailed-list-item">
+      <div className="location-weather__detailed-list-item-time">{time}</div>
+      <img
+        src={`${API_SYMBOL_URL_PREFIX}${detailedData.symbol}${API_SYMBOL_URL_POSTFIX}`}
+        alt={detailedData.time}
+        className="location-weather__detailed-list-item-symbol-img"
+      />
+      <div className="location-weather__detailed-list-item-temperature">
+        {formatTemperature(detailedData.temperature) + '°'}
```
anton-vlasik: ES6 strings please and everywhere else

src/constants/constants.js
```diff
@@ -9,12 +9,20 @@ export const API_AUTH_PASS = process.env.API_AUTH_PASS;
 export const API_BASE_URL = '/api'; //proxed to 'https://pfa.foreca.com' via webpack
 export const API_AUTH_URL = '/authorize/token';
 export const API_SEARCH_URL = '/api/v1/location/search/';
+export const API_GET_LOCATION_INFO_URL = '/api/v1/location/';
 export const API_CURRENT_WEATHER_URL = '/api/v1/current/';
 export const API_FORECAST_URL = '/api/v1/forecast/';
-export const API_FORECAST_3_HOURLY = '3hourly/';
+export const API_FORECAST_DETAILED = '3hourly/';
+export const API_FORECAST_DETAILED_PERIODS = '56';
 export const API_FORECAST_DAILY = 'daily/';
```
anton-vlasik: Api forecast daily ...what? Cucumbers?

src/constants/constants.js
```diff
@@ -9,12 +9,20 @@ export const API_AUTH_PASS = process.env.API_AUTH_PASS;
 export const API_BASE_URL = '/api'; //proxed to 'https://pfa.foreca.com' via webpack
 export const API_AUTH_URL = '/authorize/token';
 export const API_SEARCH_URL = '/api/v1/location/search/';
+export const API_GET_LOCATION_INFO_URL = '/api/v1/location/';
 export const API_CURRENT_WEATHER_URL = '/api/v1/current/';
 export const API_FORECAST_URL = '/api/v1/forecast/';
-export const API_FORECAST_3_HOURLY = '3hourly/';
+export const API_FORECAST_DETAILED = '3hourly/';
```
anton-vlasik: api forecast daily ...what? puppies? 

would it be beneficial if you would have 
```
getHourlyForecastURL= hoursNumber => `${hoursNumber}hourly/`;
```
?

src/utils/utils.js
```diff
@@ -0,0 +1,3 @@
+export function formatTemperature(temperature) {
+  return +temperature <= 0 ? temperature : `+${temperature}`;
```
anton-vlasik: Try to avoid these funky type castings like '+'. Use Number constructor or parseInt please

src/components/LocationWeather/LocationWeatherCurrent/LocationWeatherCurrent.jsx
alvikalvik: Changed null to empty string.
I was just needed some value that React will not show. I did not know that empty string is better than null.
Please explain why it's better? - because of performance or because it visually more clear?

src/components/LocationWeather/LocationWeatherDailyList/LocationWeatherDailyList.jsx
alvikalvik: I see. Changed. Also I'll use inline mappings in similar cases.

src/components/LocationWeather/LocationWeatherCurrent/LocationWeatherCurrent.jsx
anton-vlasik: Mostly because I think it is consistent to match types of initial values to values themselves. 
number -> 0
string -> ''
boolean -> true/false
object -> null / {}
array -> null / []

src/components/LocationWeather/LocationWeatherDailyListItem/LocationWeatherDailyListItem.jsx
alvikalvik: changed. I'll keep this in mind. Thanks for repeating

src/components/LocationWeather/LocationWeatherDailyListItem/LocationWeatherDailyListItem.jsx
alvikalvik: indeed. It's much better. Changed.

src/components/LocationWeather/LocationWeatherDailyListItem/LocationWeatherDailyListItem.jsx
```diff
@@ -0,0 +1,39 @@
...
+        </div>
+        <div className="location-weather__daily-list-item-minmax-temperature">{' ... '}</div>
+        <div className="location-weather__daily-list-item-minmax-temperature">
+          {formatTemperature(dayData.maxTemp) + '°'}
```
ximet: es6 string

src/constants/constants.js
```diff
@@ -33,4 +45,15 @@ export const DATA_TYPES = {
   }
 };
 
+export const WIND_DIRECTIONS_TO_IMG = {
+  N: 'w0',
```
ximet, line 49: better use full naming

src/components/LocationWeather/LocationWeatherDailyListItem/LocationWeatherDailyListItem.jsx
alvikalvik: I will keep trying... Changed (also in similar places). Thanks

src/components/LocationWeather/LocationWeather.jsx
```diff
@@ -0,0 +1,85 @@
...
+    (async () => {
+      try {
+        await weatherAPI.getToken(API_AUTH_USERNAME, API_AUTH_PASS);
+        const currentLocationWeather = await weatherAPI.getCurrentWeather(API_KIEV_ID);
+        this.setState({
+          currentLocationWeather,
+          activeDayDate: currentLocationWeather.time.split('T')[0]
+        });
+        const currentLocationInfo = await weatherAPI.getLocationInfo(API_KIEV_ID);
+        this.setState({ currentLocationInfo });
```
ximet: what if getLocationInfo === null? (catch case)

src/constants/constants.js
```diff
@@ -33,4 +45,15 @@ export const DATA_TYPES = {
...
+  NE: 'w45',
+  E: 'w90',
+  SE: 'w135',
+  S: 'w180',
+  SW: 'w225',
+  W: 'w270',
+  NW: 'w315'
+};
+
 export const API_KIEV_ID = 100703448;
```
ximet, line 59: are you sure? What if ID changed?

src/components/LocationWeather/LocationWeatherCurrent/LocationWeatherCurrent.jsx
```diff
@@ -0,0 +1,66 @@
...
+  API_SYMBOL_URL_POSTFIX,
+  API_SYMBOL_URL_PREFIX,
+  API_WIND_IMG_URL_POSTFIX,
+  API_WIND_IMG_URL_PREFIX,
+  WIND_DIRECTIONS_TO_IMG
+} from '../../../constants/constants';
+import { formatTemperature } from '../../../utils/utils';
+import './LocationWeatherCurrent.css';
+
+function LocationWeatherCurrent({ currentLocationWeather, currentLocationInfo }) {
```
ximet: strange naming. currentLocationWeather better, but I don't understand why Current? Why not previous or next? Not obvious

src/components/LocationWeather/LocationWeatherDetailedList/LocationWeatherDetailedList.jsx
alvikalvik: I see. Changed this to inline. Also understood about filterData naming.

src/components/LocationWeather/LocationWeatherDetailedListItem/LocationWeatherDetailedListItem.jsx
alvikalvik: After your comment realized that it's not enough reliable. So I decided to use js Date objects and manipulate with them. In any case the server will send something that can be converted into Date object.

src/components/LocationWeather/LocationWeatherDetailedListItem/LocationWeatherDetailedListItem.jsx
alvikalvik: sure, thanks. Changed all places where found

src/constants/constants.js
alvikalvik: not Cucumbers :) Changed to API_FORECAST_DAILY_ENDPOINT and API_FORECAST_DETAILED_ENDPOINT

src/constants/constants.js
alvikalvik: In this case no. We'll loose at least in network traffic.
API has two fixed endpoints '3hourly/' and 'hourly/'. There are no `${OTHER_NUMBER}hourly/` endpoints on server.
In theory I can implement getAnyHourly function on base of 'hourly/' endpoint - get all hours data and filter only needed hours. But in this case we'll seriously loose in network traffic because the largest part of received data will be filtered.
So my point is to use the current implementation with '3hourly/' endpoint because it gives exactly needed amount of data.

src/utils/utils.js
alvikalvik: ok, understood

src/components/LocationWeather/LocationWeatherDailyListItem/LocationWeatherDailyListItem.jsx
alvikalvik: changed

src/constants/constants.js
alvikalvik: Is it better even if the sever gives values in this format?
This object is created for mapping received from server values to image names.

src/components/LocationWeather/LocationWeather.jsx
alvikalvik: For now it shows empty strings to UI and Error message to console. In future I'm planning to create a flag in state that shows unsuccess data fetching from server and depending on this flag show some error message to user.
I think to implement this with migrating to Redux.

src/constants/constants.js
alvikalvik: Good question. I need your advice with this.
Now I hardcoded Kiev Id because search on foreca gives 4 results for "Kiev".
Is it good approach if I'll search "Kiev" on foreca server, take the item from search result where item.name === "Kiev", and take id of this exact matching item?
If this is a bad idea please advise a better way to do this.
Thanks.

src/components/LocationWeather/LocationWeatherCurrent/LocationWeatherCurrent.jsx
alvikalvik: Sorry, did not fully catch this comment. "Current weather" is synonym of "Now weather" or "weather at the moment". "Current" is not only position in array (previous, current, next), it's also "now, nowadays, presently, right now..."

src/components/LocationWeather/LocationWeatherCurrent/LocationWeatherCurrent.jsx
alvikalvik: Thanks

src/constants/constants.js
anton-vlasik: ok that's fine. I just thought that maybe they have more flexible endpoints for that purpose.

src/components/LocationWeather/LocationWeatherDetailedList/LocationWeatherDetailedList.jsx
```diff
@@ -0,0 +1,24 @@
...
+
+import './LocationWeatherDetailedList.css';
+
+function LocationWeatherDetailedList({ currentLocationDetailedWeather, activeDayDate }) {
+  return (
+    <ul className="location-weather__detailed-list">
+      {currentLocationDetailedWeather
+        ? currentLocationDetailedWeather
+            .filter(
+              detailedData => new Date(detailedData.time).setHours(0, 0, 0, 0) === activeDayDate
```
anton-vlasik, line 11: could you elaborate a bit on this logic? what do you try to filter out and what to leave?

src/components/LocationWeather/LocationWeatherDetailedListItem/LocationWeatherDetailedListItem.jsx
```diff
@@ -0,0 +1,45 @@
...
+      <img
+        src={`${API_SYMBOL_URL_PREFIX}${detailedData.symbol}${API_SYMBOL_URL_POSTFIX}`}
+        alt={detailedData.time}
+        className="location-weather__detailed-list-item-symbol-img"
+      />
+      <div className="location-weather__detailed-list-item-temperature">
+        {`${formatTemperature(detailedData.temperature)}°`}
+      </div>
+      <img
+        src={`
```
anton-vlasik: strange formatting? is it manual ?

src/utils/utils.js
```diff
@@ -0,0 +1,3 @@
+export function formatTemperature(temperature) {
+  return Number(temperature) <= 0 ? temperature : `+${temperature}`;
```
anton-vlasik, line 2: are negative temperature values returned with minus?

src/components/LocationWeather/LocationWeatherDetailedList/LocationWeatherDetailedList.jsx
alvikalvik: 1.	I receive following data array from server. It’s data with all intervals during nearest week. 

 
DetailedWeatherData

2.	User can select interesting for him or her day (in code it is called activeDayDate). It can be any of the nearest 7 days.
3.	I filter from the array items with active day in time field. Time of the day is does not matter. 
For example, user wants to get detailed forecast for tomorrow. I need to filter all items where time field contains "2021-10-08"
4.	I operate with js Data objects for more reliability. So for each item I convert item.time to Date format without hours, minutes, seconds, milliseconds: new Date(detailedData.time).setHours(0, 0, 0, 0)
5.	If the converted item.time equals to activeDayDate it goes to filtered array.

src/components/LocationWeather/LocationWeatherDetailedListItem/LocationWeatherDetailedListItem.jsx
alvikalvik: If you about these strings spacing:
`${API_WIND_IMG_URL_PREFIX}${
                WIND_DIRECTIONS_TO_IMG[detailedData.windDirString]
             }${API_WIND_IMG_URL_POSTFIX}`

it's automatic formatting with prettier vscode plugin. 

If you about other formatting - please explain what do you mean.
Thanks

src/utils/utils.js
alvikalvik: Yes. Negative temperatures with '-'. I tested with real fetched data - zero, negative and positive temperatures are properly displayed by UI

src/constants/constants.js
ximet: ok, got it

src/components/LocationWeather/LocationWeatherCurrent/LocationWeatherCurrent.jsx
ximet: yup, could you describe you naming (LocationWeatherCurrent). And why current - last word?

alvikalvik: I was trying to show that LocationWeatherCurrent component should be a child of LocationWeather component. I wanted to make its name similar to LocationWeatherDailyList  and LocationWeatherDetailedList.
Now I understand that LocationWeatherCurrent name is not obvious.

How do you advise to name subcomponent of LocationWeather that shows current weather?
Thanks.

src/constants/constants.js
alvikalvik: I added geo location detection via navigator.geolocation. If for some reason geo detection does not work API_DEFAULT_ID constant will be used.

Now geo detection is temporary inside LocationWeather component, with migration to redux it will be moved.

src/components/LocationWeather/LocationWeatherCurrent/LocationWeatherCurrent.jsx
alvikalvik: The component renamed to LocationWeatherCurrentInfo meaning that it's current info about location weather.
Thanks for your patience :)

anton-vlasik (approved):

ximet (approved):

PR #70, opened 2021-10-08
Prop types

alvikalvik:
Typization of components with prop-types

### DIFF
--- a/src/components/LocationWeather/LocationWeatherDailyList/LocationWeatherDailyList.jsx
+++ b/src/components/LocationWeather/LocationWeatherDailyList/LocationWeatherDailyList.jsx
@@ -1,5 +1,5 @@
 import PropTypes from 'prop-types';
-import { currentLocationDailyWeatherType } from '../../../types/types';
+import { CurrentLocationDailyWeatherType } from '../../../types/types';
 import LocationWeatherDailyListItem from '../LocationWeatherDailyListItem/LocationWeatherDailyListItem';
 
 import './LocationWeatherDailyList.css';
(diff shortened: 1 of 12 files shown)


src/components/LocationWeather/LocationWeatherDailyList/LocationWeatherDailyList.jsx
```diff
@@ -25,4 +27,10 @@ function LocationWeatherDailyList({
   );
 }
 
+LocationWeatherDailyList.propTypes = {
+  currentLocationDailyWeather: currentLocationDailyWeatherType,
+  activeDayDate: PropTypes.number, //timestamp
```
anton-vlasik, line 32: are you sure it is a number?  in milliseconds

src/types/types.js
```diff
@@ -0,0 +1,58 @@
+import PropTypes, { number, string } from 'prop-types';
+
+export const currentLocationInfoType = PropTypes.shape({
```
anton-vlasik: I'd suggest to name types starting with Capital letter

src/components/LocationWeather/LocationWeatherDailyList/LocationWeatherDailyList.jsx
alvikalvik: yes, number in  milliseconds, I checked one more time.
I need it as a number instead of Data object because in future I'll put activeDayDate into redux store.

src/types/types.js
alvikalvik: Renamed the type names to starting with Capital letter

anton-vlasik (approved):

ximet (approved):

PR #72, opened 2021-10-08
Add Search input

JaneEllison:
Add Search input, - with styles, liveSearching.
No Redux at this time.


src/components/SearchInput/SearchInput.jsx
```diff
@@ -1,7 +1,62 @@
...
+import { dataService } from '../../services/dataService';
 
 function SearchInput() {
-  return <div>Search input</div>;
+  const [searchQuery, setSearchQuery] = useState('');
+  const [searchResults, setSearchResults] = useState([]);
+
+  useEffect(() => {
+    if (searchQuery) {
+      searchCities();
```
anton-vlasik: again I would suggest to pass search query as a param here. What if you want to move searchCities out of this component's scope

src/components/SearchInput/SearchInput.jsx
```diff
@@ -1,7 +1,62 @@
...
+    }
+  }, [searchQuery]);
```
anton-vlasik: Also as an improvement - think how can you optimize your code. Right now on every key stroke you are sending a request. Could you **debounce** it?

JaneEllison: Done

src/components/SearchInput/SearchInput.jsx
JaneEllison: It's true, corrected

src/components/SearchInput/SearchInput.jsx
```diff
@@ -1,7 +1,63 @@
...
+    } catch (error) {
+      console.log(error);
+    }
+  }
+
+  const handleChange = e => {
+    setSearchQuery(e.target.value);
+  };
+
+  const renderSearchResults = searchResults.map(item => (
```
anton-vlasik: I don't think there is a lot of added value to move this mapping here. It is super usual to write it "inline" right n JSX

src/components/SearchInput/SearchInput.jsx
```diff
@@ -1,7 +1,63 @@
...
+    <li key={item.id} className={styles.searchResultItem}>
+      <img src={SearchLocationIcon} alt="location" />
+      <span>{item.name}</span>
+    </li>
+  ));
+
+  return (
+    <div className={styles.searchWrapper}>
+      <div className={styles.searchForm}>
+        <DebounceInput
```
anton-vlasik: I want you to write it yourself, please. I allow you to use debounce function from lodash.debounce. But Debounce input should be written by yourself. Don't forget to remove unneeded packages after that

src/components/SearchInput/SearchInput.jsx
JaneEllison: ok, got it

src/components/SearchInput/SearchInput.jsx
JaneEllison: i did it

src/hooks/useDebounce.jsx
```diff
@@ -0,0 +1,16 @@
+import { useState, useEffect } from 'react';
+
+export function useDebounce(value, delay) {
+  const [dbValue, setDbValue] = useState('');
+
+  useEffect(() => {
+    const handler = setTimeout(() => {
```
anton-vlasik: This is more of a timeout Id.

src/components/SearchInput/SearchInput.jsx
```diff
@@ -1,21 +1,22 @@
...
 import SearchIcon from '../../assets/images/search-icon.png';
 import SearchLocationIcon from '../../assets/images/search-location-icon.png';
-import { NUMBER_OF_RESULTS } from '../../constants/forecaApi';
-import { dataService } from '../../services/dataService';
 
 function SearchInput() {
   const [searchQuery, setSearchQuery] = useState('');
   const [searchResults, setSearchResults] = useState([]);
 
+  const debouncedSearchQuery = useCallback(useDebounce(searchQuery, 1000));
```
anton-vlasik: useCallback accepts a function... and a list of dependencies, if you don't want it to run every time returning you new value. In your case you are passing a string to it from what I see. So I'm a bit confused.

src/hooks/useDebounce.jsx
JaneEllison: Ok, fixed

src/components/SearchInput/SearchInput.jsx
JaneEllison: Yes. I rewrote function and forget to remove this one. Fixed.

anton-vlasik (approved):

ximet:
resolve conflict

ximet (approved):

PR #73, opened 2021-10-11
Redux setup

alvikalvik:
Initial Redux setup - folders structure created, app wrapped with provider,  1 reducer and 1 container added


src/actionTypes/index.js
```diff
@@ -0,0 +1,4 @@
+export const SET_IS_TOKEN_RECEIVED = 'SET_IS_TOKEN_RECEIVED';
+export const SET_TOKEN_EXPIRATION_TIME = 'SET_TOKEN_EXPIRATION_TIME';
+export const SET_IS_FETCHING_IN_PROGRESS = 'SET_IS_FETCHING_IN_PROGRESS';
+export const SET_IS_FETCHING_ERROR = 'SET_IS_FETCHING_ERROR';
```
anton-vlasik: SET_FETCHING_ERROR ?

src/actions/ServerApiActions.js
```diff
@@ -0,0 +1,26 @@
+import {
+  SET_IS_FETCHING_ERROR,
+  SET_IS_FETCHING_IN_PROGRESS,
+  SET_IS_TOKEN_RECEIVED,
+  SET_TOKEN_EXPIRATION_TIME
+} from '../actionTypes';
+
+export const setIsTokenReceived = isTokenReceived => ({
+  type: SET_IS_TOKEN_RECEIVED,
+  payload: isTokenReceived
```
anton-vlasik: If you go with payload, then I'd suggest to make it an object, so that for other cases you would have an ability to pass more data

src/actionTypes/index.js
alvikalvik: Sorry, I did not catch the question:
1. If you ask to explain the logic of SET_IS_FETCHING_ERROR action - it sets isFetchingError flag to true or false value. This flag shows were any errors during data fetching (I'll set this flag if some of axios requests will be rejected)
2. If you mean that SET_FETCHING_ERROR  name is better than SET_IS_FETCHING_ERROR in this case - please write, I'll rename.

src/actions/ServerApiActions.js
alvikalvik: All the payloads converted to objects. Thanks

src/actionTypes/index.js
anton-vlasik: I'd say there is not much useful info for error in a boolean

alvikalvik: I finally got.
Replaced the boolean flag with entire Error object. Now if the error happens all possible info about it will be available.
Thanks for the explanation, now it's much clear what is useful to store.

anton-vlasik (approved):

ximet (approved):

PR #74, opened 2021-10-11
Setup redux

DmitriyKrivovyaz:


client/src/redux/reducers/index.js
```diff
@@ -0,0 +1,12 @@
+import { combineReducers } from 'redux';
+import { dailyWeatherReducer } from './dailyWeatherReducer';
+import { hourlyWeatherReducer } from './hourlyReducer';
+import { currentWeatherReducer } from './currentWeatherReducer';
+import { locationReducer } from './locationReducer';
+
+export const rootReducer = combineReducers({
+  dailyWeatherReducer,
```
anton-vlasik: as I was saying when you name things like that  - you will have them named like that in your store, which is I think semantically is not the best thing

DmitriyKrivovyaz: Thanks. Changed

anton-vlasik (approved):

ximet (approved):

PR #77, opened 2021-10-11
Main header

usermeme:


src/app/layouts/private-layout/components/header/Header.tsx
```diff
@@ -1,3 +1,43 @@
-const Header: React.FC = () => null;
+import * as React from 'react';
+import { shallowEqual } from 'react-redux';
+import * as S from './styles';
+import * as Components from 'components';
+import { useSelector, useTranslation } from 'hooks';
+import { RootState } from 'store/types';
+import { hedaerSelectors } from './duck';
```
ximet: hedaer? maybe header?

src/app/layouts/private-layout/components/header/Header.tsx
```diff
@@ -1,3 +1,43 @@
...
+
+const Header: React.FC = () => {
+  const { t } = useTranslation();
```
ximet, line 10: I don't like naming with one letter.

usermeme: hook useTranslation from i18n return function that name is "t",  also i18n has type TFunction for that function, I can use like this `const { t: translation } = `, but I think this is unnecessary, since function "t" and hook useTranslation always go in pairs. 

What do you think?

src/app/layouts/private-layout/components/header/Header.tsx
usermeme: fixed

src/app/layouts/private-layout/components/header/Header.tsx
ximet: ok, got it

AV29 (approved):

ximet (approved):

PR #85, opened 2021-10-13
Add current date

JaneEllison:
Add current current date block with date and time.


src/components/CurrentDate/CurrentDate.jsx
```diff
@@ -1,10 +1,29 @@
...
 import styles from './CurrentDate.module.scss';
 import { FORECAST_PATHS } from '../../constants/forecaApi';
+import { updatingDateInterval } from '../../constants/date';
+import { getFormattedDate } from '../../utils/getFormattedDate';
 
 function CurrentDate({ weatherImg }) {
+  const [currentDate, setCurrentDate] = useState(getFormattedDate);
+
+  useEffect(() => {
+    let timerId = setInterval(() => {
```
anton-vlasik, line 13: I'd suggest to refactor a bit: 
```
useEffect(() => {
    let timerId = setInterval(() => {
      setCurrentDate(getFormattedDate());
    }, updatingDateInterval);

    return () => {
      clearInterval(timerId);
    };
  }, [currentDate]);
```
1) Optional, but I think code is still pretty readable
2) I'd include necessary dependency in a list so that if you want to introduce another state in this component later - you won't reinitialize your timer due to unrelated cause


P.S. If you are using setInterval not setTimeout - would it work if you initialize and destroy in ONCE ?

JaneEllison: Understand. Thanks. 
Yes, it would work. 
And with setTimeout it wouldn't work.

anton-vlasik (approved):

ximet (approved):

PR #91, opened 2021-10-15
Location weather via redux

alvikalvik:


src/actions/CurrentLocationActions.js
```diff
@@ -0,0 +1,97 @@
...
+
+export const setCurrentLocationDetailedWeather = detailedWeather => ({
+  type: SET_CURRENT_LOCATION_DETAILED_WEATHER,
+  payload: {
+    detailedWeather
+  }
+});
+
+export const getLocationDataById = locationId => async (dispatch, getState) => {
+  if (getState().serverApi.isTokenReceived && !getState().serverApi.isFetchingInProgress) {
```
anton-vlasik: getState could be cached

src/actions/CurrentLocationActions.js
```diff
@@ -0,0 +1,97 @@
...
+    dispatch(setIsFetchingInProgress(true));
+
+    try {
+      const currentLocationInfo = weatherAPI.getLocationInfo(locationId);
```
anton-vlasik: it's not currentLocationInfo, currentLocationWeather or currentLocationDailyWeather or currentLocationDetailedWeather - those are promises. Deferred objects. Bad naming

src/actions/CurrentLocationActions.js
```diff
@@ -0,0 +1,97 @@
...
+      );
+      const currentLocationDetailedWeather = weatherAPI.getForecast(
+        API_FORECAST_DETAILED_ENDPOINT,
+        locationId,
+        {
+          periods: API_FORECAST_DETAILED_PERIODS
+        }
+      );
+
+      const results = await Promise.all([
```
anton-vlasik: Are you sure you handle exceptional situations here?

anton-vlasik: you sure Promise.all is a good thing here? One fails - everything is failed.

src/actions/CurrentLocationActions.js
```diff
@@ -0,0 +1,97 @@
...
+        currentLocationDailyWeather,
+        currentLocationDetailedWeather
+      ]);
+
+      const [
+        currentLocationInfoResult,
+        currentLocationWeatherResult,
+        currentLocationDailyWeatherResult,
+        currentLocationDetailedWeatherResult
+      ] = results;
```
anton-vlasik: you can destructure it on line 63 directly

src/actions/CurrentLocationActions.js
```diff
@@ -0,0 +1,97 @@
...
+      dispatch(setCurrentLocationInfo(currentLocationInfoResult));
+      dispatch(setCurrentLocationWeather(currentLocationWeatherResult));
+      dispatch(setCurrentLocationDailyWeather(currentLocationDailyWeatherResult));
+      dispatch(setCurrentLocationDetailedWeather(currentLocationDetailedWeatherResult));
+    } catch (error) {
+      dispatch(setFetchingError(error));
+      console.log(error);
+    }
+
+    dispatch(setIsFetchingInProgress(false));
```
anton-vlasik: This code is most likely unreachable. I'd better use **finally** block

src/actions/CurrentLocationActions.js
alvikalvik: Thanks, I created a variable with getState() value, so the function is called once only.

src/actions/CurrentLocationActions.js
alvikalvik: Indeed. Thanks, renamed all these promises with Promise postfix.

src/actions/CurrentLocationActions.js
alvikalvik: Yes, I tested all the promises inside Promise.all separately with incorrect params and all the errors were catched.

alvikalvik: it's logical necessity in this case. I have to be sure that all these promises are resolved. In other case it's possible that only part of data will be updated.
For example it's senseless to show detailed weather without daily weather, or show weather without city name.
Or second example - current location changes from Kiev to Minsk. In this case I have to fetch all data about Minsk, I can't just update part of data because the old data is about Kiev.

src/actions/CurrentLocationActions.js
alvikalvik: Thanks, done

src/actions/CurrentLocationActions.js
alvikalvik: Indeed. Thanks. I was rewriting .then.catch.finally to try/catch with async/await and forgot about the finally block.
Changed.

anton-vlasik (approved):

ximet (approved):

PR #96, opened 2021-10-18
Add theme switcher to redux

JaneEllison:
Refactor switching themes. Add themes to redux.

### DIFF
--- a/src/layouts/Header/ThemeSwitcher/Container.jsx
+++ b/src/layouts/Header/ThemeSwitcher/Container.jsx
@@ -9,13 +9,10 @@ import ThemeSwitcher from './ThemeSwitcher';
 const ThemeSwitcherContainer = ({ theme, switchTheme }) => {
   function changeTheme() {
     const currentTheme = theme === THEMES.light ? THEMES.dark : THEMES.light;
+    localStorage.setItem('theme', currentTheme);
 
     switchTheme(currentTheme);
   }
-
-  useEffect(() => {
-    localStorage.setItem('theme', theme);
-  }, [theme]);
 
   return <ThemeSwitcher theme={theme} onSwitch={changeTheme} />;
 };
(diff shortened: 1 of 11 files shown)


anton-vlasik (approved):

src/layouts/Header/ThemeSwitcher/Container.jsx
```diff
@@ -0,0 +1,38 @@
...
+
+const ThemeSwitcherContainer = ({ theme, switchTheme }) => {
+  function changeTheme() {
+    const currentTheme = theme === THEMES.light ? THEMES.dark : THEMES.light;
+
+    switchTheme(currentTheme);
+  }
+
+  useEffect(() => {
+    localStorage.setItem('theme', theme);
```
ximet: How it works?

src/redux/reducers/themeReducer.js
```diff
@@ -1,16 +1,18 @@
-import { THEMES } from '../../constants/themes';
 import { SET_CURRENT_THEME } from '../types/themeTypes';
 
+import { THEMES } from '../../constants/themes';
+import { localStorageService } from '../../services/localStorageService';
+
 const INITIAL_STATE = {
-  theme: THEMES.light
+  currentTheme: localStorageService.getCurrentTheme() || THEMES.light
```
ximet, line 7: initial state from localStorage? Why?

src/layouts/Header/ThemeSwitcher/Container.jsx
JaneEllison: If you change theme, theme name save to local storage, and then if you refresh page you see this one which you choose earlier

src/redux/reducers/themeReducer.js
JaneEllison: If the user refresh the page, state will set saved theme value. 
Or I should do this in the component?

const currentTheme=  localStorageService.getCurrentTheme();

 useEffect(() => {
    switchTheme(currentTheme)
  }, []);

Something like this?

src/layouts/Header/ThemeSwitcher/Container.jsx
JaneEllison: I moved setting data to localStorageService, removed useEffect and calling the function on changeTheme function.

ximet (approved):

PR #98, opened 2021-10-18
Selected locations via redux

alvikalvik:


src/components/SelectedLocations/SelectedLocationsList/SelectedLocationsList.jsx
```diff
@@ -0,0 +1,34 @@
...
+        ? selectedLocations.map(locationData => {
+            return (
+              <SelectedLocationsListItem
+                key={locationData.id}
+                locationData={locationData}
+                deleteSelectedLocation={deleteSelectedLocation}
+              />
+            );
+          })
+        : ''}
```
anton-vlasik: 1) selectedLocations ? ... : null
2) or selectedLocations && selectedlocations.map ... 

P.S. do you need <ul> wrapper if there are no locations?

src/components/SelectedLocations/SelectedLocationsListItem/SelectedLocationsListItem.jsx
```diff
@@ -0,0 +1,61 @@
...
+      <button
+        className="selected-locations__list-item-remove-btn"
+        type="button"
+        onClick={() => deleteSelectedLocation(id)}
+      >
+        Remove
+      </button>
+    </li>
+  ) : (
+    ''
```
anton-vlasik: null

src/hooks/initialize.js
```diff
@@ -22,5 +34,11 @@ export function useInitialize() {
     ) {
       dispatch(getToken());
     }
-  }, [isTokenReceived, tokenExpirationTime, fetchingError, dispatch]);
+  }, [
+    isTokenReceived,
+    tokenExpirationTime,
+    tokenExpirationTime <= Number(new Date()),
```
anton-vlasik: can you elaborate on that?

src/reducers/selectedLocationsReducer.js
```diff
@@ -0,0 +1,27 @@
...
+  DELETE_SELECTED_LOCATION,
+  CLEAR_SELECTED_LOCATIONS
+} from '../actionTypes';
+
+const initialState = [];
+
+const selectedLocationsReducer = (state = initialState, action) => {
+  switch (action.type) {
+    case PUT_SELECTED_LOCATION:
+      if (state.some(data => data.id === action.payload.id)) {
```
anton-vlasik: OPTIONAL: but I personally against logic branching in reducers. Reducers should be better as much atomic as possible. You can create additional action instead

src/components/SelectedLocations/SelectedLocationsList/SelectedLocationsList.jsx
alvikalvik: thanks, && syntax looks much better. I'll use it.
Also, you are right - the wrapper is not needed if there are no locations. I moved the wrapper inside condition

src/components/SelectedLocations/SelectedLocationsListItem/SelectedLocationsListItem.jsx
alvikalvik: Also changed to && syntax, thanks again

src/hooks/initialize.js
alvikalvik: This is replaced with timer as we discussed in chat

src/reducers/selectedLocationsReducer.js
alvikalvik: Replaced this with 2 atomic reducers.
Thanks.

alvikalvik:
I refactored timers inside weather components. Now the timers are maximally simple and independent.

src/actions/SelectedLocationsActions.js
```diff
@@ -0,0 +1,75 @@
...
+  }
+};
+
+export const updateAllSelectedLocationsData = () => async (dispatch, getState) => {
+  const state = getState();
+  if (state.serverApi.isTokenReceived && !state.serverApi.isFetchingInProgress) {
+    try {
+      dispatch(setIsFetchingInProgress(true));
+
+      state.selectedLocations.forEach(async location => {
```
anton-vlasik: feel a bit iffy about this code... meaning awaiting inside the loop. How about doing all together (pseudo code)?

```
const weatherRequests = state.selectedLocations.map(({ id })=> weatherAPI.getCurrentWeather(id));
const locationRequests = state.selectedLocations.map(({ id }) => weatherAPI. getLocationInfo(id));

const locationInfo = await Promise.all[locationRequests];
const weatherInfo = await Promise.all[weatherRequests];

dispatch(putSelectedLocations(...));

```

?

alvikalvik: You are right. Your approach is more obvious and the info appears on the screen faster.
Refactored.
Thanks.

ximet (approved):

src/actions/SelectedLocationsActions.js
```diff
@@ -0,0 +1,77 @@
...
+        weatherAPI.getLocationInfo(id)
+      );
+      const locationWeatherPromises = state.selectedLocations.map(({ id }) =>
+        weatherAPI.getCurrentWeather(id)
+      );
+
+      const locationInfos = await Promise.all(locationInfoPromises);
+      const weatherInfos = await Promise.all(locationWeatherPromises);
+
+      locationInfos.forEach((item, index) => {
```
anton-vlasik, line 58: Optionally you could create a single action and put all the data with a single dispatch

src/components/SelectedLocations/SelectedLocations.jsx
```diff
@@ -0,0 +1,31 @@
+import './SelectedLocations.css';
+import SelectedLocationsList from './SelectedLocationsList/SelectedLocationsList';
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+
+import { SelectedLocationType } from '../../types/types';
+
+class SelectedLocations extends PureComponent {
```
anton-vlasik, line 8: Why PureComponent?

anton-vlasik (approved):

src/actions/SelectedLocationsActions.js
alvikalvik: In general I understand how to do do this (and for what) but let me record this and do it in future if the time will left after implementing of other critical features. Thanks.

src/components/SelectedLocations/SelectedLocations.jsx
alvikalvik: I briefly heard about PureComponent from you on one of early lections in zoom. From this very short explanation I thought that PureComponent is almost the same as Component but it optimizes rerender when props and state are changeless. So I thought that PureComponent  is preferable.
After your question I understand that it's actually wrong position. Please explain (here or in general chat) when we should use PureComponent and when Component.
Thanks a lot.

PR #99, opened 2021-10-19
Add weather to redux

JaneEllison:
Refactor weather components. Add getting data to redux.


src/layouts/CityForecast/Container.jsx
```diff
@@ -0,0 +1,26 @@
+import { connect } from 'react-redux';
+import PropTypes from 'prop-types';
+
+import { THEMES, BG_IMAGES } from '../../constants/themes';
+import { CityForecastTypes, CityInfoTypes } from '../../types/WeatherDataTypes';
+import CityForecast from './CityForecast';
+
+function CityForecastContainer({ cityForecast, cityInfo, theme }) {
+  const themeBg = theme === THEMES.light ? BG_IMAGES.light : BG_IMAGES.dark;
```
anton-vlasik: I'd make it a part of inner component in order to keep Container just a wrapper over the CityForecast without any additional logic. You can use selector for that matter. Plus, themeBg - is it a single source property always applicable for all components across the app? If yes - it can be shared from a higher level

src/layouts/DetailedForecast/HourlyForecast/Container.jsx
```diff
@@ -0,0 +1,22 @@
+import { connect } from 'react-redux';
+
+import { getCorrectHourlyData } from '../../../utils/hourlyChartSettings';
+import { HourlyCityForecastTypes } from '../../../types/WeatherDataTypes';
+
+import HourlyForecast from './HourlyForecast';
+
+function HourlyForecastContainer({ hourlyForecast }) {
+  const correctHourlyData = getCorrectHourlyData(hourlyForecast);
```
anton-vlasik: same here. Use selector

src/layouts/CityForecast/Container.jsx
JaneEllison: I added selectors. 
themeBg is not for all components, it is just background image for this components based on choosen theme

src/layouts/DetailedForecast/HourlyForecast/Container.jsx
JaneEllison: Done

src/types/weatherDataTypes.js
```diff
@@ -56,3 +56,9 @@ export const DailyCityForecastTypes = PropTypes.arrayOf(
     windDir: number
   })
 );
+
+export const HourlyChartData = PropTypes.shape({
```
ximet: move it to component

src/types/weatherDataTypes.js
```diff
@@ -56,3 +56,9 @@ export const DailyCityForecastTypes = PropTypes.arrayOf(
     windDir: number
```
ximet: check all file

src/types/weatherDataTypes.js
JaneEllison: moved

src/types/weatherDataTypes.js
JaneEllison: I checked and refactored all about propTypes. 
Also I created selectors for some components data

src/redux/reducers/weatherReducer.js
```diff
@@ -20,6 +23,25 @@ const INITIAL_STATE = {
...
+      return {
+        ...state,
+        isFetching: false
+      };
+
+    case FETCH_WEATHER_FAILURE:
+      return {
+        ...state,
+        isFetching: false,
+        isFetchingFailure: true
```
anton-vlasik: Is there anywhere an action to reset isFetchingFailure to false back again...after an error occurred? Plus, I suggest to work with actual error object, because then you can manipulate with meaningful data, codes, messages inside. Not only boolean fact that error has just occurred

src/redux/selectors/locationSelectors.js
```diff
@@ -0,0 +1,5 @@
+export const getCityInfo = state => {
```
anton-vlasik: `export const getCityInfo = state => state.location.currentLocation`?

src/redux/selectors/weatherSelectors.js
```diff
@@ -0,0 +1,14 @@
+import { createSelector } from 'reselect';
+import { getFormattedHourlyData } from '../../utils/hourlyChartSettings';
+
+const gethourlyCityForecast = state => state.weather.hourlyCityForecast;
+
+export const getCityForecast = state => {
```
anton-vlasik: `export const getCityForecast = state => state. weather. cityForecast` ?

src/redux/selectors/weatherSelectors.js
```diff
@@ -0,0 +1,14 @@
...
+  const { relHumidity, symbol, symbolPhrase, temperature, windSpeed } = state.weather.cityForecast;
+
+  return { relHumidity, symbol, symbolPhrase, temperature, windSpeed };
+};
+
+export const getHourlyChartData = createSelector([gethourlyCityForecast], forecast =>
```
anton-vlasik, line 8: OPTIONAL: 
`const getHourlyChartData = createSelector([gethourlyCityForecast], getFormattedHourlyData); `

src/views/WeatherForecast/Container.jsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+
+import { connect } from 'react-redux';
+import { CURRENT_CITY_ID } from '../../constants/forecaApi';
+import { getWeatherInfo } from '../../redux/actions/weatherActions';
+
+import WeatherForecast from './WeatherForecast';
+
+class WeatherForecastContainer extends React.PureComponent {
```
anton-vlasik: Could, please, you explain why are you using PureComponent here ?

src/views/WeatherForecast/Container.jsx
```diff
@@ -0,0 +1,40 @@
...
+    this.props.getWeatherInfo(CURRENT_CITY_ID);
+  }
+
+  render() {
+    if (this.props.isFetching) {
+      return <span> Loading...</span>;
+    }
+
+    if (this.props.isFetchingFailure) {
+      return <span> Something went wrong...</span>;
```
anton-vlasik: What exactly went wrong?

src/redux/selectors/weatherSelectors.js
JaneEllison: Fixed.

src/redux/selectors/locationSelectors.js
JaneEllison: Fixed.

src/redux/actions/weatherActions.js
```diff
@@ -1,20 +1,57 @@
...
+const setHourlyCityForecast = forecast => ({
   type: SET_HOURLY_CITY_FORECAST,
   payload: forecast
 });
+
+export const getWeatherInfo = location => async dispatch => {
+  dispatch(fetchWeatherStart);
+
+  try {
+    await dataService.getForecastToken();
```
alvikalvik, line 44: I'm not sure but ask. Do you need token only for weather? I think for some other actions it also will be needed (for example  location search). Maybe it's worth to make separate and independent from other actions token getting?

src/redux/selectors/weatherSelectors.js
JaneEllison: I will remember it for the future, but for now I will leave it for a better understanding

src/views/WeatherForecast/Container.jsx
JaneEllison: Oh. Most likely, I forgot to remove it at the beginning, when I watched what it does and how it works, and then, due to what I transferred by copying, everything remained the same.
There is no need for a PureComponent, as we use it when we want to inline add a componentShouldUpdate. And if the parent props or state is changed, then the child changed too. And the PureComponent allows you to avoid this if its property stay the same or linked to thesame object

src/redux/actions/weatherActions.js
JaneEllison: Later i planned make something like initialize action where i get token and then all data than i needed. And if we wanna search location token already will be on dataService object

src/redux/reducers/weatherReducer.js
JaneEllison: fixed. I created error message field and there I cat put error text wich I will show to user

src/views/WeatherForecast/Container.jsx
JaneEllison: something...)
Fixed.

src/redux/actions/weatherActions.js
```diff
@@ -1,20 +1,58 @@
...
+    const { cityForecast, dailyCityForecast, hourlyCityForecast } =
+      await dataService.getFullForecast(location);
+
+    dispatch(setCityForecast(cityForecast));
+    dispatch(setDailyCityForecast(dailyCityForecast));
+    dispatch(setHourlyCityForecast(hourlyCityForecast));
+
+    dispatch(fetchWeatherSuccess);
+  } catch (error) {
+    dispatch(fetchWeatherFailure(FETCHING_ERROR_TEXT));
```
anton-vlasik: dispatch(fetchWeatherFailure(error)) or something error based

ximet (approved):

src/redux/actions/weatherActions.js
JaneEllison: fixed

anton-vlasik (approved):

PR #104, opened 2021-10-21
Make theme selector

glebdronchenko:


src/utils/prepareData.js
```diff
@@ -12,4 +12,17 @@ function prepareChartData(hourlyForecast) {
   return [hours, temperatures];
 }
 
-export { prepareChartData };
+function prepareThemesList(themeConfig) {
```
anton-vlasik: If you already using lodash - there is 'lodash/values' I think

src/components/ThemeSelect/components/ThemeDropDown/ThemeDropDown.jsx
```diff
@@ -0,0 +1,18 @@
+import ThemeItem from '../ThemeItem/ThemeItem';
+import classes from './ThemeDropDown.module.scss';
+import themeConfig from '../../../../config/themeConfig';
+import { prepareThemesList } from '../../../../utils/prepareData';
+
+function ThemeDropDown() {
+  const themeList = prepareThemesList(themeConfig);
```
anton-vlasik: Again: this is out of scope of ThemeDropDown. Apparently you can do that once in your app. Or even change your theme object structure to avoid taking values of themeList object

src/components/ThemeSelect/components/ThemeItem/ThemeItem.jsx
```diff
@@ -0,0 +1,21 @@
+import classes from './ThemeItem.module.scss';
+import ThemeContext from '../../../../providers/themeContext';
+
+function ThemeItem({ theme }) {
+  return (
+    <ThemeContext.Consumer>
```
anton-vlasik: Are you sure that Context.Consumer should wrap each individual Item but not the whole list of items instead?

src/hooks/commonHooks.js
```diff
@@ -0,0 +1,22 @@
...
+    }
+  };
+
+  useEffect(() => {
+    document.addEventListener('click', handleClick, true);
+
+    return () => {
+      document.removeEventListener('click', handleClick, true);
+    };
+  }, [refElement]);
```
anton-vlasik: why do you use refElement as a dependency?

src/hooks/commonHooks.js
```diff
@@ -0,0 +1,22 @@
+import { useEffect, useRef, useState } from 'react';
+
+export function useDropdownVisible(initialVisible) {
```
anton-vlasik: Seems like you want to check outside clicks of your element. It is not a matter of visibility  - you are inspecting if click was outside or inside. Moreover is it only working with dropdowns? So it seems to me that the name is not quite good enough. Either you work only with dropdowns or it is more like useClickoutside. - but then it is up to consuming component to do something when click was outside. Task of this hook would just to tell the outter world that it was registered outside passed element,

src/utils/prepareData.js
glebdronchenko: No, I don't use lodash.

src/hooks/commonHooks.js
glebdronchenko: No. It's my mistake. It doesn't have to be here. At first I wanted to provide this variable to hook as function parameter and I thought, that this parameter could change but I was wrong.

src/hooks/commonHooks.js
glebdronchenko: I agree it isn't correct hook name.

src/components/ThemeSelect/components/ThemeDropDown/ThemeDropDown.jsx
glebdronchenko: I have moved all theme config methods to theme config.

src/components/ThemeSelect/components/ThemeItem/ThemeItem.jsx
glebdronchenko: I have removed this theme context and pass context values ​​from parent component.

src/components/ThemeSelect/components/ThemeDropDown/ThemeDropDown.jsx
```diff
@@ -0,0 +1,23 @@
+import ThemeItem from '../ThemeItem/ThemeItem';
+import classes from './ThemeDropDown.module.scss';
+import { getThemesList } from '../../../../config/themeConfig';
+import ThemeContext from '../../../../providers/themeContext';
+
+function ThemeDropDown({ currentTheme, selectTheme }) {
+  const themeList = getThemesList();
```
anton-vlasik: Again: does it need to be in the scope of this component? HINT: no

glebdronchenko: I moved it outside of component scope.

ximet (approved):

anton-vlasik (approved):

PR #107, opened 2021-10-21
Locations page with routing

alvikalvik:


src/components/LocationWeather/LocationWeather.jsx
```diff
@@ -31,6 +31,7 @@ class LocationWeather extends PureComponent {
         <LocationWeatherCurrentInfo
           currentLocationWeather={this.props.currentLocationWeather}
           currentLocationInfo={this.props.currentLocationInfo}
+          addToSelectedLocations={this.props.addToSelectedLocations}
```
anton-vlasik, line 34: how about removal? would make more sense to have `onChangeLocationSelection={this.props....}` ?

src/components/LocationsSearch/LocationsSearch.jsx
```diff
@@ -0,0 +1,52 @@
...
+function LocationsSearch({ isReadyForSearch, searchResults, clearSearch, searchLocations }) {
+  const [inputValue, setInputValue] = useState('');
+  const debouncedSearchLocations = debounce(searchLocations, DEBOUNCE_WAIT_TIME);
+
+  function handleInputChange(event) {
+    const value = event.target.value;
+    setInputValue(value);
+
+    if (value === '') {
+      clearSearch();
```
anton-vlasik, line 19: can you elaborate why do you need to clear something out if value itself is already empty?

src/components/LocationsSearch/LocationsSearch.jsx
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    if (value.length >= MIN_CHARACKTERS_FOR_SEARCH) {
```
anton-vlasik: check spelling: characters

src/components/LocationWeather/LocationWeatherCurrentInfo/LocationWeatherCurrentInfo.css
```diff
@@ -11,6 +11,12 @@
   padding: 0;
 }
 
+.location-weather__add-btn {
+  margin-top: 0.2rem;
+  font-size: 0.8rem;
+  font-weight: 700;
```
JaneEllison: it will be better to move values such as font-size and font-weight to constants maybe

src/components/LocationsSearch/LocationsSearch.css
```diff
@@ -0,0 +1,22 @@
...
+  padding: 0;
+}
+
+.locations-search__input {
+  margin-top: 0.5rem;
+}
+
+.locations-search__input-description {
+  margin-top: 0.2rem;
+  font-size: 0.7rem;
```
JaneEllison: and here

src/components/LocationsSearch/LocationsSearchReuslts/LocationsSearchReuslts.css
```diff
@@ -0,0 +1,34 @@
...
+  padding: 0.3rem;
+  color: var(--primary-color);
+  text-decoration: none;
+  font-size: 1rem;
+}
+
+.locations-search__no-results {
+  margin: 0.5rem;
+  font-size: 0.8rem;
+  font-weight: 700;
```
JaneEllison: here too

src/components/LocationWeather/LocationWeather.jsx
alvikalvik: Removal from selected locations is implemented in other component - SelectedLocations.
Here addToSelectedLocations is needed if user wants to add currently viewing location to selected.
I do not think that it worth to add the removal here. However if it will be needed in future - it's simple to add.

src/components/LocationsSearch/LocationsSearch.jsx
alvikalvik: The logics is following:
We need to clear search results if user clears search input.
Let's imagine that user searched something previously - the results of this search are currently shown.
If user wants to start a new search he or she deletes all text in the search input (users often do this before new search) - results should disappear at this moment. In other case user will continue looking at the old results thinking that they are new...

src/components/LocationsSearch/LocationsSearch.jsx
alvikalvik: corrected, thanks

src/components/LocationWeather/LocationWeatherCurrentInfo/LocationWeatherCurrentInfo.css
alvikalvik: Thanks in advise. I use css codestyle with rems for this project. Possibly I'll move font-sizes and font-weights to constants in future if time will left after critical features implementation. Thanks one more time.

src/components/LocationWeather/LocationWeatherCurrentInfo/LocationWeatherCurrentInfo.jsx
```diff
@@ -60,6 +68,20 @@ function LocationWeatherCurrentInfo({ currentLocationWeather, currentLocationInf
...
+        <button
+          className="location-weather__add-btn"
+          type="button"
+          onClick={() => {
+            addToSelectedLocations({
+              id: String(locationId),
+              locationInfo: currentLocationInfo,
+              locationWeather: currentLocationWeather
+            });
+          }}
```
glebdronchenko: I think that it would be better, if you passed here variable instead of literal object.

src/components/LocationsSearch/LocationsSearch.jsx
glebdronchenko: > The logics is following: We need to clear search results if user clears search input. Let's imagine that user searched something previously - the results of this search are currently shown. If user wants to start a new search he or she deletes all text in the search input (users often do this before new search) - results should disappear at this moment. In other case user will continue looking at the old results thinking that they are new...

You need to use this function because you send search request after value.length >= MIN_CHARACTERS_FOR_SEARCH. If you didn't use this condition, you wouldn't use this function.

src/components/LocationWeather/LocationWeatherCurrentInfo/LocationWeatherCurrentInfo.jsx
alvikalvik: Can you please explain in more details? For what we need to create additional variable? I think the literal object is readable enough in this case.

src/components/LocationWeather/LocationWeatherCurrentInfo/LocationWeatherCurrentInfo.jsx
```diff
@@ -42,11 +50,11 @@ function LocationWeatherCurrentInfo({ currentLocationWeather, currentLocationInf
...
-          alt={`${currentLocationWeather.windDirString} ${currentLocationWeather.windSpeed} m/s`}
+          alt={`${currentLocationWeather.windDirString} ${currentLocationWeather.windSpeed} ${WIND_SPEED_TEXT}`}
           className="location-weather__current-weather-wing-img"
         />
         <div className="location-weather__current-weather-wind-value">
-          {currentLocationWeather.windSpeed} m/s
+          {currentLocationWeather.windSpeed} {WIND_SPEED_TEXT}
         </div>
       </div>
     </>
```
glebdronchenko, line 65: This block would be more readable, if you moved it to separate component.

src/components/LocationWeather/LocationWeatherCurrentInfo/LocationWeatherCurrentInfo.jsx
glebdronchenko: I think would be better place all variable in one place (above than return), because it will be more readable.

src/components/LocationsSearch/LocationsSearch.jsx
alvikalvik: > You need to use this function because you send search request after value.length >= MIN_CHARACTERS_FOR_SEARCH. If you didn't use this condition, you wouldn't use this function.

It's senseless to remove this condition because it will cause excessive network load (Server answers only if the condition is true). Also it's weird to send request with empty query.

src/components/LocationWeather/LocationWeatherCurrentInfo/LocationWeatherCurrentInfo.jsx
alvikalvik: ok, I see. Refactored. Thanks.

src/components/LocationsSearch/LocationsSearch.jsx
glebdronchenko: > > You need to use this function because you send search request after value.length >= MIN_CHARACTERS_FOR_SEARCH. If you didn't use this condition, you wouldn't use this function.
> 
> It's senseless to remove this condition because it will cause excessive network load (Server answers only if the condition is true). Also it's weird to send request with empty query.

I think you can use another methods for load reduction (for example debounce). 
Server answer to you without error if you send query string of any length. And I think that you need to have just one truth.
In this situation you check this condition and the server check the same condition too. It's my opinion =) Maybe I'm wrong))

src/components/LocationWeather/LocationWeatherCurrentInfo/LocationWeatherCurrentInfo.jsx
alvikalvik: Most likely this block will not be reused so I prefer to not separate it.

src/components/LocationsSearch/LocationsSearch.jsx
alvikalvik: We do not want to enlarge user waiting in any case so it's much faster to check the condition on client's side than send server request and wait for it (with slow 3G it can take several seconds). So I still think that I need this condition here.
Also debounce is already used in couple with the condition check.
Thanks a lot.

src/components/LocationWeather/LocationWeatherCurrentInfo/LocationWeatherCurrentInfo.css
ximet: why not today?

alvikalvik: Today is ok. Yesterday I had doubts according terms so had to move this task. Now I changed ccs files included in this PR. Other css files will be changed in following separate PR (I decided to not include them into this PR for not making it too large).
Thanks.

anton-vlasik (approved):

PR #108, opened 2021-10-22
About page with router

alvikalvik:


package.json
```diff
@@ -21,6 +21,7 @@
     "axios": "0.21.4",
     "core-js": "3.2.1",
     "history": "4.10.1",
+    "lodash": "^4.17.21",
```
ximet: for what lodash? I didn't see anything import from lodash

alvikalvik: I use debounce from lodash in previous PR. In this PR I just reinstalled lodash (it seems it was installed improperly first time, I received warning during one of builds).

anton-vlasik: you can install individual packages: in this case lodash.debounce

alvikalvik: Great, thanks. Replaced lodash  with lodash.debounce

ximet (approved):

PR #109, opened 2021-10-22
Add location to redux

JaneEllison:
Refactor location components. Add changing location data to redux. Add recent cities in redux.


src/services/dataService.js
```diff
@@ -27,11 +27,13 @@ export const dataService = {
     const cityForecast = await this.getCurrentForecast(id);
     const dailyCityForecast = await this.getForecast(FORECAST_TYPES.day, id);
     const hourlyCityForecast = await this.getForecast(FORECAST_TYPES.hour, id);
+    const cityInfo = await this.getCityInfo(id);
```
alvikalvik: Do you really need to start these promises one by one and await previous before next start? If no you can start them in parallel with Promise.all() This approach should save time.

JaneEllison: Hmm. You're right. I changed it.

src/App.jsx
```diff
@@ -6,8 +7,13 @@ import './assets/styles/constants/theme.css';
 import Header from './Layouts/Header/Header';
 import Main from './layouts/Main/Main';
 import Footer from './layouts/Footer/Footer';
+import { initApp } from './redux/actions/initAppActions';
```
anton-vlasik, line 10: you are not supposed to import it like that. You need to use connect. Otherwise you are tidily bind your view to a specific app initializer. It is not idiomatic from react-redux standpoint too.

src/layouts/DetailedForecast/FavoriteCities/Container.jsx
```diff
@@ -0,0 +1,14 @@
+import { connect } from 'react-redux';
+import FavoriteCities from './FavoriteCities';
+import { getCityInfo } from '../../../redux/selectors/locationSelectors';
+
+const mapStateToProps = state => {
+  const { id, name, country } = getCityInfo(state);
+
+  return {
+    favoriteCities: state.location.recentCities,
+    currentCity: { id, name, country }
```
anton-vlasik: `currentCity: getCityInfo(state);`?

src/App.jsx
JaneEllison: But I used connect there on 32-38 line

src/layouts/DetailedForecast/FavoriteCities/Container.jsx
JaneEllison: i wanna get only this fields for this component, shoul I hand over all object fields?

JaneEllison: Fixed. I created selectors

src/App.jsx
anton-vlasik: Sorry, missed that

anton-vlasik (approved):

PR #110, opened 2021-10-25
Setting dark theme.

glebdronchenko:


src/utils/chartSettings.js
```diff
@@ -17,8 +17,14 @@ const hourlyLineChartOptions = {
   scales: {
     y: {
       ticks: {
+        color: '#00b2ff',
```
alvikalvik: Color hardcoded here? Maybe it worth to move it to constants?

glebdronchenko: I think that you right. Because when we will need to change application color scheme it will be enough to change color into one place. I didn't it because I thought that this settings just for one chart and it won't use at another place.
Thanks a lot!

anton-vlasik (approved):

PR #111, opened 2021-10-25
Make contacts template

glebdronchenko:


src/components/Contacts/Contacts.module.scss
```diff
@@ -0,0 +1,47 @@
...
+  padding: 30px 40px;
+  margin: 0 10px;
+  display: flex;
+  align-items: center;
+
+  .contactsWrapper {
+    width: 100%;
+    .title {
+      font-size: $contacts-title-font-size;
+      font-weight: bold;
```
JaneEllison: i'm not sure, but maybe better to create a constants for weight and use numeric values

src/components/Contacts/Contacts.module.scss
```diff
@@ -0,0 +1,47 @@
...
+
+    .contact:not(:last-child) {
+      margin-bottom: 20px;
+    }
+
+    .contactEmail {
+      text-align: center;
+
+      .email {
+        font-weight: bold;
```
JaneEllison: and here too

src/components/FeedbackForm/FeedbackForm.module.scss
```diff
@@ -0,0 +1,130 @@
+@import '../../assets/styles/constants.scss';
+
+$title-font-size: 18px;
```
JaneEllison: why do you use separate local constants here?

src/components/FeedbackForm/FeedbackForm.jsx
```diff
@@ -0,0 +1,104 @@
...
+    <div className={classes.formWrapper}>
+      <h3 className={classes.title}>Feedback form</h3>
+      {isSended ? (
+        <div className={classes.result}>
+          <h3 className={classes.title}>Thanks for your message!</h3>
+          <div className={classes.description}>Our support team will answer you later.</div>
+          <button className={classes.btn} onClick={() => setIsSended(false)}>
+            Return to form
+          </button>
+        </div>
```
DmitriyKrivovyaz, line 29: I would suggest to create a formField component, cox u reuse this fields.

src/components/FeedbackForm/FeedbackForm.module.scss
DmitriyKrivovyaz: Agree. I think it is not useful to use a const here it is the same as u would write font-size: 18px in this file

glebdronchenko: I agree with you. In this case this const don't have to be here.

src/components/FeedbackForm/FeedbackForm.jsx
glebdronchenko: I wanted it, but I couldn't do it in this PR. There are a lot of changed files.

src/components/Contacts/Contacts.module.scss
glebdronchenko: I think you right that it has to be as a const. In this case if we need to change font or font weight, we will do it at one place.
But I don't think so it will be better if I will use it as numeric value. 
I have changed this values in in files of this PR and have created new PR where this value was setting at all project files.
https://github.com/ximet/ReactTask/pull/113

src/utils/validationForm.js
```diff
@@ -0,0 +1,43 @@
+export function validateEmailField(value) {
+  const validate = [];
```
anton-vlasik: validate is a verb

src/utils/validationForm.js
```diff
@@ -0,0 +1,43 @@
...
+
+function getErrors(validate) {
+  let result;
+  const errors = validate.filter(validate => validate !== undefined);
+  if (errors.length > 0) result = errors.join(', ');
+
+  return result;
+}
+
+function checkValidateEmail(value) {
```
anton-vlasik: validateEmail would be enough for this name I guess

src/utils/validationForm.js
```diff
@@ -0,0 +1,43 @@
...
+  const regExp = new RegExp(rule, 'g');
+
+  if (!regExp.test(value)) {
+    error = 'Invalid email address';
+  }
+
+  return error;
+}
+
+function checkValidateRequire(value) {
```
anton-vlasik: validateRequired

src/config/contacts.js
```diff
@@ -0,0 +1,37 @@
...
+      code: 'social_tw',
+      icon: iconTw,
+      link: 'https://twitter.com/',
+      target: '__blank'
+    }
+  }
+};
+
+export function getContactEmail() {
+  return contactsConfig.email;
```
anton-vlasik: Can't you just use contactsConfig.email? Why do you need this getter?

src/config/contacts.js
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export function getSocialsList() {
+  const socials = contactsConfig.socials;
```
anton-vlasik: Why can't you just keep your data in desired shape?

src/config/contacts.js
glebdronchenko: I have removed this getter. 
I thought that it will be better than take access to the object field directly.

src/config/contacts.js
glebdronchenko: In this case yes, I can use data in desired shape. But I think, that map will be better, when we need to get specific element of list. In map we can use key for getting some element and in the same time it don't use loop for searching this element.

I have changed this data shape.

src/utils/validationForm.js
glebdronchenko: I have changed it.

src/utils/validationForm.js
glebdronchenko: I have changed it.

src/utils/validationForm.js
glebdronchenko: I have changed it.

anton-vlasik (approved):

PR #119, opened 2021-10-25
Refactoring search request

glebdronchenko:


src/hooks/searchHooks.js
```diff
@@ -0,0 +1,20 @@
...
+      const { data } = await ApiService.getLocationsSearch(searchString);
+      responseData = data.locations;
+    } catch (error) {
+      console.error(error);
+    }
+
+    setLocations(responseData);
+  }, [searchString]);
+
+  return [locations];
```
alvikalvik: Please explain why return [locations], not return locations? For instance initially locations [], so we'll return [[]]. Should it be so?

anton-vlasik (approved):

PR #121, opened 2021-10-26
Add Slider

JaneEllison:
Add slider component. Add slider for Favorite cities, daily forecast components. Add changing current city by Favorite cities card click. Fix styles.


src/components/Slider/Slider.module.scss
```diff
@@ -0,0 +1,67 @@
...
+    justify-content: center;
+    position: absolute;
+    right: -50px;
+    height: 50px;
+    top: 50%;
+    transform: translate(0, -50%);
+    width: 50px;
+    border: 0;
+    z-index: 1;
+    font-size: 34px;
```
glebdronchenko: Hardcoded value.

src/components/Slider/Slider.module.scss
```diff
@@ -0,0 +1,67 @@
...
+    background-color: $primary-bg-color;
+    color: $arrow-btn-color;
+    cursor: pointer;
+    outline: none;
+    transition: font-size $default-transition-delay ease;
+
+    &:hover {
+      color: $hover-arrow-btn-color;
+      font-size: 40px;
```
glebdronchenko: Hardcoded value.

src/components/Slider/Slider.module.scss
JaneEllison: Thank you. Fixed

src/components/Slider/Slider.module.scss
JaneEllison: Thank you. Fixed

src/components/SearchInput/SearchInput.module.scss
```diff
@@ -90,6 +91,7 @@
 
 .search-wrapper:hover .search-results {
   opacity: 1;
+  z-index: 15;
```
anton-vlasik: Why -5, 15 - how do you decide what values to assign?

src/redux/selectors/locationSelectors.js
```diff
@@ -1,9 +1,18 @@
...
 export const getCityInfo = state => state.location.currentCity;
+const getFavoriteCities = state => state.location.recentCities;
 
 export const getCityTitle = createSelector([getCityInfo], ({ country, name, id }) => ({
   country,
   name,
   id
 }));
+
+export const getCityCards = createSelector([getFavoriteCities], favoriteCities => {
```
anton-vlasik: this seems to be a selector and I suggested you to name them differently from simple getters

src/redux/selectors/locationSelectors.js
```diff
@@ -1,9 +1,18 @@
...
+  return favoriteCities.map(city => ({
+    id: city.id,
+    slide: <CityCard info={city} />
```
anton-vlasik: This time of mapping usually just happen in render function, so it is either useless or - did you just mean to 
```
export const getCityCards = createSelector(getFavoriteCities, favoriteCities => {
     return favoriteCities.map(city => ({ id: city.id, slideInfo: city }));
});
```
which then again makes it pretty useless as city already contains id

src/redux/selectors/weatherSelectors.js
```diff
@@ -18,3 +20,10 @@ export const getShortCityForecast = createSelector(
     windSpeed
   })
 );
+
+export const getDailyForecastCards = createSelector([getDailyForecast], dailyForecast => {
+  return dailyForecast.map(forecast => ({
+    id: forecast.date,
+    slide: <WeatherCard weatherInfo={forecast} />
```
anton-vlasik: same here.

src/components/SearchInput/SearchInput.module.scss
JaneEllison: I rewiew all z-indexes and remove excess.  
What about value step  - I am thought about situation when we need to add element with z-index between this two value. And in this case we need to rewrite all indexes and add new index. And if we have value step equal to 5, we can add any value between. but these are situations that rarely happend)

src/redux/selectors/locationSelectors.js
JaneEllison: renamed to selectCityCards

src/redux/selectors/locationSelectors.js
JaneEllison: I create an array with CityCard components and pass this array to Slider component where i render those components, thats why i need id field. But I refactored this piece for better understanding - rename property name, remove id property  - now I passed only array of components. And also I added library for random ids in Slider component.

src/redux/selectors/weatherSelectors.js
JaneEllison: I refactored it.

anton-vlasik (approved):

PR #124, opened 2021-10-26
Setting preloaders

glebdronchenko:


src/selectors/selectorsFavorite.js
```diff
@@ -1,3 +1,7 @@
 export const selectFavoriteCityIds = favoriteCitiesList => {
```
anton-vlasik: These are not selectors in a sense of memoization or supporting data from state. These are just simple helpers.

src/actions/locationsManagerActions.js
```diff
@@ -131,9 +116,46 @@ export const changeFavoriteLocationsList =
...
+  };
+
+export const getFavoriteForecast =
+  (locationId: string): ThunkActionCachedForecasts =>
+  async (dispatch: DispatchCachedForecasts, getState: GetStoreState): Promise<void> => {
+    dispatch(changeFavoriteForecastState(locationId, false));
+    try {
+      const { data } = await ApiService.getCurrentForecast(locationId);
+      const locationForecast = {
+        cacheTimeStamp: getCurrentTime(),
```
anton-vlasik, line 144: Setting cachTeimStamp every time seems tedious to me. Could you elaborate more why do you need to pass it every time?

src/actions/locationsManagerActions.js
```diff
@@ -131,9 +116,46 @@ export const changeFavoriteLocationsList =
...
+    } catch (error) {
+      console.error(error);
+    }
+    dispatch(changeFavoriteForecastState(locationId, true));
+  };
+
 export const setCurrentHourlyForecast =
   (locationId: number | string): ThunkActionHourlyForecast =>
   async (dispatch: DispatchHourlyForecast, getState: GetStoreState): Promise<void> => {
+    dispatch(changeHourlyForecastState(false));
```
anton-vlasik: Do you change state of your hourly forecast? and how exactly do you change it? It is totally obscure from the given name

src/actions/locationsManagerActions.js
```diff
@@ -157,6 +181,7 @@ export const setCurrentDailyForecast =
     } catch (error) {
       console.error(error);
     }
+    dispatch(changeDailyForecastState(true));
```
anton-vlasik: same here and in every other place

src/actions/locationsManagerActions.js
glebdronchenko: This action run, just if we didn't have cached data in state or if cached time is over.

src/actions/locationsManagerActions.js
glebdronchenko: Yes, it isn't correct name. I have renamed all preloader actions.

src/selectors/selectorsFavorite.js
glebdronchenko: I have rewrite this function and have used reselect library for it.

src/actions/locationsManagerActions.js
glebdronchenko: I have changed this function name.

src/actions/locationsManagerActions.js
```diff
@@ -207,4 +233,5 @@ export const setWarnings =
     } catch (error) {
       console.error(error);
     }
+    dispatch(toggleCurrrentLocationPreloader(true));
```
anton-vlasik: too much of an "r".

src/actions/locationsManagerActions.js
```diff
@@ -157,6 +181,7 @@ export const setCurrentDailyForecast =
     } catch (error) {
       console.error(error);
     }
+    dispatch(toggleDailyForecastPreloader(true));
```
anton-vlasik: Toggle means that it will switch the state to inverted respecting previous. Doesn't it? But here you are setting the value. See in other similar places too

glebdronchenko: I have renamed this function and another function from preloader actions. I have changed prefix "toggle" to "changeVisible"

src/actions/locationsManagerActions.js
glebdronchenko: Done.

anton-vlasik (approved):

PR #125, opened 2021-10-26
Refactor getting default locations data

JaneEllison:


src/redux/actions/initAppActions.js
```diff
@@ -38,13 +39,21 @@ export const getAllData = location => async dispatch => {
...
   const currenLocationFromGeo = await locationService.getCurrentLocation();
-  const defaultLocation = getState().location.currentCity.id;
+  const defaultLocation = DEFAULT_CURRENT_CITY_ID;
   const location = currenLocationFromGeo || defaultLocation;
 
   await dataService.getForecastToken();
+
+  const cityInfoPromises = DEFAULT_RECENT_CITIES_ID.map(id => dataService.getCityInfo(id));
+
+  const cityInfos = await Promise.all(cityInfoPromises);
```
anton-vlasik: TBH I doubt that info has plural form

JaneEllison: Corrected to "citiesInfo"

anton-vlasik (approved):

src/redux/actions/initAppActions.js
```diff
@@ -38,13 +39,21 @@ export const getAllData = location => async dispatch => {
...
+  const defaultLocation = DEFAULT_CURRENT_CITY_ID;
   const location = currenLocationFromGeo || defaultLocation;
 
   await dataService.getForecastToken();
+
+  const cityInfoPromises = DEFAULT_RECENT_CITIES_ID.map(id => dataService.getCityInfo(id));
+
+  const citiesInfo = await Promise.all(cityInfoPromises);
+  citiesInfo.forEach(info => {
+    dispatch(setRecentCity(info));
```
anton-vlasik, line 55: could you set all info all together?

anton-vlasik (changes requested):
you need to resolve conflicts

src/redux/actions/initAppActions.js
JaneEllison: do you mean do something like this: 

 `dispatch(setRecentCity(citiesInfo ));` ?

no, i could't because when I set new City I call function which check is this sity already in recentCities state and if I pass an array there i need to pass array of one city later to add new city, and loop over this array to check is this city already exist

anton-vlasik: It would make things more optimised though. To loop through array in your function vs. dispatch actions in a loop.

anton-vlasik (approved):

PR #131, opened 2021-10-27
Make testing

glebdronchenko:


src/selectors/selectorsForecast.js
```diff
@@ -1,5 +1,6 @@
 import { getCurrentTime } from '../utils/dateTimeUtils';
 
 export const selectCurrentForecast = (forecasts, locationId) => {
-  return forecasts[locationId]?.forecast;
+  const forecast = forecasts[locationId]?.forecast;
```
anton-vlasik: 1) `return forecasts[locationId]?.forecast || {}` ?
2) this is not a selector per se. It is neither memoizing anything nor working with redux state

glebdronchenko: Yes. It's just helper. I have understood. I thought, that it will be better put these helpers and selectors in one place.

I have changed this function.

anton-vlasik (approved):

PR #137, opened 2021-10-29
Add tests

DmitriyKrivovyaz:


client/src/redux/actions/locationAction.test.js
```diff
@@ -0,0 +1,27 @@
...
+  it('Should return SET_LOCATION action', () => {
+    const location = {
+      id: 1,
+      name: 'name',
+      country: 'country',
+      lon: 1,
+      lat: 1
+    };
+
+    const expectResult = {
```
alvikalvik: I think variable name expectedResult is better than expectResult in this case because expectResult sounds like function name.

ximet (approved):

PR #149, opened 2021-12-08
Feature/redux toolkit statement

Ksenia-Tulikova:
I've added base structure of redux statement.
You could find a sample action for page AboutUs. Its just a test action, dont pay on it much of your attention.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -20,6 +20,7 @@
   "dependencies": {
     "@mui/icons-material": "^5.2.0",
     "@mui/material": "^5.2.1",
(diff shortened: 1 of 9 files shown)


client/package.json
```diff
@@ -20,6 +20,7 @@
   "dependencies": {
     "@mui/icons-material": "^5.2.0",
     "@mui/material": "^5.2.1",
+    "@reduxjs/toolkit": "^1.6.2",
```
anton-vlasik, line 23: Honestly saying I would like us to proceed without meta tools for redux in this course. Let's train with pure redux and then you can play around with helper libraries.

PR #159, opened 2022-01-26
Implement API connect and fetching data

makeychik:
There is some problems with repeating requests when app init


src/constants.js
```diff
@@ -0,0 +1,8 @@
+const proxy = 'https://immense-hamlet-20285.herokuapp.com/';
+const api = 'https://pfa.foreca.com';
+export const url = proxy + api;
+
+export const authData = {
+  user: 'nmakeychik1',
+  password: '64vwAHlJXbJN'
```
ximet: nice, right now I can use your credentials. Thank you)

src/App.jsx
```diff
@@ -1,5 +1,48 @@
...
+import SearchLocation from './components/SearchLocation';
+import { url, authData } from './constants';
+
 function App() {
-  return <div>Hello world!</div>;
+  const [token, setToken] = useState('');
+  const [coords, setCoords] = useState({});
+  const [searchedLocation, setSearchedLocation] = useState({});
+
+  navigator.geolocation.getCurrentPosition(
```
JuliaShavyrova: do you want to call the function on each render?

src/App.jsx
```diff
@@ -1,5 +1,48 @@
+import React, { useEffect, useState } from 'react'; // нужно ли импортировать в каждый файл?
```
JuliaShavyrova: please remove question comments
and yes it is ok to import in each file

src/App.jsx
```diff
@@ -1,5 +1,48 @@
...
+      setCoords(pos);
+    },
+    err => {
+      console.warn(`ERROR(${err.code}): ${err.message}`);
+    }
+  );
+
+  const isSearchedLocationEmpty = Object.keys(searchedLocation).length === 0;
+
+  const getSearchedLocation = location => {
```
JuliaShavyrova: as i can see, your functions getSearchedLocation and getLocation don't really get anything, please rename

src/constants.js
```diff
@@ -0,0 +1,8 @@
+const proxy = 'https://immense-hamlet-20285.herokuapp.com/';
```
JuliaShavyrova: please make you own proxy as discussed

src/constants.js
makeychik: moved them to .env on the server side

src/App.jsx
makeychik: no, refactored this moment

src/App.jsx
makeychik: done

src/App.jsx
makeychik: I renamed them but not sure about good naming. Need to research best practicies to name callbacks which pass as props

src/constants.js
makeychik: done

JuliaShavyrova (approved):

PR #160, opened 2022-01-26
Api

DamnD2:
Added weather API and getting data based on location. Data output to the main page.


DamnD2:
merge conflicts resolved

src/api/weatherApi.js
```diff
@@ -0,0 +1,31 @@
+const user = 'vitalya_daineko';
+const password = 'TXr1Bm3ShDik';
```
ximet: nice, right now I can use your credentials. Thank you)

src/api/weatherApi.js
```diff
@@ -0,0 +1,31 @@
+import { user, password } from '../../env';
+
+const proxy = 'https://morning-cove-62020.herokuapp.com/';
```
ximet: please, create custom proxy

src/components/Header/Header.jsx
```diff
@@ -1,20 +1,44 @@
...
+
   return (
     <div className="header">
       <Link className="header__logo" to="/">WeatherApp</Link>
       <nav className="menu">
         <ul className="menu__list">
           <li className="menu__list-item">
-            <Link className="menu__list-link" to="/">Home</Link>
+            <Link
+              className={`menu__list-link ${hash === '/' ? 'active' : ''}`}
```
JuliaShavyrova: as you don't use css modules, what if you'll have "active" class name in other components? please use BEM completely or don't use it at all

src/api/weatherApi.js
DamnD2: https://github.com/DamnD2/proxy

src/api/weatherApi.js
```diff
@@ -0,0 +1,31 @@
+import { USER, PASSWORD } from '../../env';
```
JuliaShavyrova: as far as your env.js file not included in project directory (because of gitignore rules), anyone who will grab your code now, will see compilation error like "env not found". please fix it (google about  .env and how it works)

JuliaShavyrova (approved):

PR #168, opened 2022-02-08
Added new pages

DamnD2:
added a page for the list of available cities with the current weather, as well as a page with detailed weather information for the selected city


src/components/SelectedCityPage/SelectedCityPage.jsx
```diff
@@ -0,0 +1,52 @@
...
+    const location = `${cityData.lon},${cityData.lat}`;
+    const token = getCookie('token');
+
+    weatherApi.getLocationInfo(location, token)
+      .then((data) => setLocationInfo(data));
+
+    weatherApi.getCurrentWeather(location, token)
+      .then((data) => setCurrentWeather(data));
+
+    weatherApi.getTodaysWeather(location, token)
```
JuliaShavyrova: can you get token value inside your request function?

src/core/Launcher.jsx
```diff
@@ -1,16 +1,21 @@
 import React, { StrictMode, useState, useMemo } from 'react';
 import PropTypes from 'prop-types';
 
-import { SnackbarContext } from './contexts';
+import { SnackbarContext, WorldWeatherContext } from './contexts';
 
 function Launcher({ children }) {
   const [snackbar, setSnackbar] = useState({ isOpen: false, message: '' });
   const memoizedSnackbar = useMemo(() => [snackbar, setSnackbar], [snackbar]);
```
JuliaShavyrova: why do you use useMemo here?

src/components/WorldWeatherPage/WorldWeatherPage.jsx
```diff
@@ -0,0 +1,60 @@
...
+        setTimeout(() => {
+          weatherApi.getLocationInfo(cityLocation, token)
+            .then((data) => {
+              weatherApi.getCurrentWeather(cityLocation, token)
+                .then((weather) => {
+                  const fullData = { ...data, ...weather };
+                  setWorldWeather((prev) => ([...prev, fullData]));
+                });
+            });
+        }, (index + 1) * 250);
```
JuliaShavyrova: if server returns you 429 error, please don't make it suffer. also your user won't be happy if network speed slows down and he needs to wait for responses from your synced requests

src/core/Launcher.jsx
DamnD2: Because the value of the state is an object. So that there would be no problems with re-rendering when changing the value to the same and the like. You can create 2 states with primitive values, but it's easier

src/components/WorldWeatherPage/WorldWeatherPage.jsx
DamnD2: In order for the server not to suffer, I need a non-existent request. Either the server developer must make the necessary rest api, or another technical task. It is impossible to accomplish the task in any other way.

JuliaShavyrova: server developer doesn't have requirements to make for you some special request. but you have API with requests that provides necessary data and you don't have a requirement to show all that data in one moment on the same page. so task is feasible but you need to think about another solution

JuliaShavyrova (approved):

PR #171, opened 2022-02-10
Feedback form

MariaMalakhovaCSI:
Feedback view:
- react-hook-form
- FeedbackForm as Survey with input fields and rating
- Save data to localStorage on submit


ximet (approved):

src/atomic-components/Star/Star.jsx
```diff
@@ -0,0 +1,15 @@
+/* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
```
JuliaShavyrova: please do not ignore eslint warnings (read more about accessibility) and use semantically correct HTML elements

PR #175, opened 2022-02-14
Routing, home page

makeychik:
- routing
- basic home page layout
- view of weather details + icon


src/components/WeatherSymbol/WeatherSymbol.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import classes from './WeatherSymbol.module.scss';
+
+const WeatherSymbol = ({ symbol, symbolPhrase }) => {
+  return (
+    <div className={classes.wrapper}>
+      <img src={require(`../../assets/${symbol}.png`)} alt={symbolPhrase} />
```
JuliaShavyrova, line 7: why do you use require here?

makeychik: I used it to be able to use local path to image in component instead of use path from index.html
found it here
https://stackoverflow.com/a/39999421 

but I think if I will delete images from src there won't be this syntax

JuliaShavyrova (approved):

PR #176, opened 2022-02-17
Add form to FeedbackPage

anastasiyaShcherenkova:


src/App.scss
```diff
@@ -29,6 +29,12 @@ a {
   text-decoration: none;
 }
 
+input, 
+textarea {
+  border: none;
+  outline: none;
```
JuliaShavyrova: please be careful when resetting "outline" prop and make sure that user without mouse can determine when he is focused on your inputs
https://developer.mozilla.org/en-US/docs/Web/CSS/outline#accessibility_concerns

src/components/feedbackForm/FeedbackForm.jsx
```diff
@@ -0,0 +1,163 @@
...
+};
+
+const ERROR_MESSAGES = {
+  isEmpty: 'This field is required',
+  minLength: 'Incorrect length',
+  maxLength: 'Incorrect length',
+  email: 'Incorrect email'
+};
+
+function useValidation(value, validators) {
```
JuliaShavyrova: just an advice - custom hooks can be moved in separate files to make your component file cleaner

JuliaShavyrova (approved):

PR #184, opened 2022-03-02
redux

DamnD2:
react context replaced with redux. ThemeSwitcher component 
connected using connect.


src/store/selectors.js
```diff
@@ -0,0 +1,4 @@
+const tokenSelector = (store) => store.token.token;
+const themeSelector = (store) => store.theme.theme;
```
JuliaShavyrova: as i see, state structure can be simplified to `{ theme: string; token: string }`

JuliaShavyrova (approved):

PR #186, opened 2022-03-09
Theme switcher

MariaMalakhovaCSI:
Add Light / Dark theme switcher
Use Foreca API to get weather icons


src/components/SearchInput/SearchInput.jsx
```diff
@@ -25,7 +25,7 @@ const formatCityTile = city => {
   return formattedTile;
 };
 
-function SearchInput({ token }) {
+function SearchInput({ token, theme }) {
```
JuliaShavyrova, line 28: just a suggestion - it would be better to get theme value from context, not passing through props. in case of using context your components can be placed anywhere in tree and won't need your parent to provide them "theme" prop (the same for token)

JuliaShavyrova (approved):

PR #196, opened 2022-03-28
Adjust search functionality

MariaMalakhovaCSI:
- Search input as form with 'No location found' condition
- Update Home view with searched and selected location (+locationId url binding)
- Get Daily forecast from API (not mocked)


src/atomic-components/LocationTile/LocationTile.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { useHistory } from 'react-router-dom';
+
+const formatCityTile = city => {
+  if (city.country === undefined) {
```
JuliaShavyrova: expression can be simplified

src/components/NavBar/NavBar.jsx
```diff
@@ -9,48 +9,52 @@ function NavBar({ token, theme, onToggleTheme }) {
...
+                  className={classes.navBar_link}
+                  activeClassName={classes.navBar_active_link}
+                >
+                  Feedback
+                </NavLink>
+              </li>
+            </ul>
 
-          <ThemeSwitcher theme={theme} onToggleTheme={onToggleTheme} />
+            <ThemeSwitcher theme={theme} onToggleTheme={onToggleTheme} />
```
JuliaShavyrova: please use context for accessing and updating theme, so every component could get "theme" value independently from their parents props

src/components/SearchInput/SearchInput.jsx
```diff
@@ -54,36 +42,41 @@ function SearchInput({ token, theme }) {
...
+
+  const handleSubmit = event => {
+    event.preventDefault();
   };
 
   return (
     <div className={classes.search_wrapper} data-theme={theme}>
-      <div className={classes.search_form}>
-        <button type="button" className={classes.search_button}>
+      <form action="" onSubmit={handleSubmit}>
```
JuliaShavyrova: please clear empty attributes

PR #206, opened 2022-05-04
added contact page with form, components(class as well) and input hook

GedasIncirauskas:
1. Created contact page with content and components inside.
2. Created simple feedback form with input components and custom hook for validation.
3. Created "star" review class component (to show that I understand how to work with class based components :) )
4. Created function to save user feedback data to local storage.
Screenshot 2022-05-04 150047
Screenshot 2022-05-04 150136
Screenshot 2022-05-04 150246


src/components/Button/Button.jsx
```diff
@@ -0,0 +1,48 @@
...
+      text = '#CC3300';
+      border = '#53585D';
+      break;
+    case 'success':
+      background = '#66BB6A';
+      text = '#FEFEFE';
+      break;
+    default:
+      background = '#0063CC';
+      text = '#FEFEFE';
```
JuliaShavyrova: please move all style logic in `.styles` file and create customized buttons according to styled-components docs and move colors to separate variables

src/components/Form/Form.jsx
```diff
@@ -0,0 +1,117 @@
...
+    inputBlurHandler: emailBlurHandler,
+    reset: resetEmail
+  } = useInput(isEmail);
+
+  const {
+    value: textareaValue,
+    isValid: textareaIsValid,
+    hasError: textareaHasError,
+    valueChangeHandler: textareaChangeHandler,
+    inputBlurHandler: textareaBlurHandler,
```
JuliaShavyrova: please rename according to logic above (what if you'll need to add another textarea to the form?)

src/components/Form/Form.jsx
```diff
@@ -0,0 +1,117 @@
...
+    reset: resetTextarea
+  } = useInput(isTextarea);
+
+  let formIsValid = false;
```
JuliaShavyrova: just a question - why is it not stored in state?

src/components/Form/Form.jsx
```diff
@@ -0,0 +1,117 @@
...
+
+  const submitHandler = event => {
+    event.preventDefault();
+
+    if (!formIsValid) {
+      return;
+    }
+
+    const usersFeedback = {
+      id: Math.floor(Math.random() * 10000) * 1,
```
JuliaShavyrova: why do you need id and what if `Math.random` generates the same number as already stored in local storage?

src/components/inputs/SimpleInput/SimpleInput.jsx
```diff
@@ -0,0 +1,22 @@
+import React, { Component } from 'react';
```
JuliaShavyrova: why component has SimpleInput file name?

src/components/Form/Form.jsx
```diff
@@ -0,0 +1,117 @@
...
+        changeHandler={nameChangeHandler}
+        blurHandler={nameBlurHandler}
+        labelName="name"
+        labelTitle="Your name"
+        errorMessage="Please enter your name"
+        isError={userNameHasError}
+      />
+      <Input
+        inputValue={emailValue}
+        changeHandler={emailChangeHandler}
```
JuliaShavyrova: why not just use internal react attributes names as prop names for event handlers?

JuliaShavyrova (approved):

src/components/Form/Form.jsx
GedasIncirauskas: I added dummy id, because usually data has id, I understood that `Math.random` could generate the same number. I will deleted it and leave without dummy ID.

src/components/Form/Form.jsx
GedasIncirauskas: I used it because if I use `useState()` I had infinity loop.

src/components/inputs/SimpleInput/SimpleInput.jsx
GedasIncirauskas: Ok, fixed naming to `Input`.

src/components/Form/Form.jsx
GedasIncirauskas: Ok, thank you. Fixed :)

src/components/Form/Form.jsx
GedasIncirauskas: I used it because it is more clear for me. I didn't want pass to input react attributes names as props and use them.
I will fixed then to usually internal react attributes names.

src/components/Button/Button.jsx
GedasIncirauskas: Fixed, I left only what is needed.

.gitignore
```diff
@@ -24,6 +24,8 @@ package-lock.json
 /build*
 /build/*
 .env
+.eslintignore
+.eslintrc.js
```
JuliaShavyrova: no need to add these files in gitignore

src/components/Form/Form.jsx
```diff
@@ -0,0 +1,109 @@
...
+      <TextArea
+        value={reviewValue}
+        onChange={reviewChangeHandler}
+        onBlur={reviewBlurHandler}
+        labelName="feedback"
+        labelTitle={translations.msg_user_review_label}
+        errorMessage={translations.msg_user_review_error}
+        isError={reviewHasError}
+      />
+      <Rate onChange={getUserRating} />
```
JuliaShavyrova, line 100: you can make `Rate` component completely [controlled](https://reactjs.org/docs/forms.html#controlled-components) so `rating` state to be the only source of truth. we will talk about it on tuesday so you can skip it for now

src/components/StarRating/StarRating.jsx
```diff
@@ -0,0 +1,59 @@
...
+
+    this.state = {
+      stars: [],
+      rating: 0,
+      hovered: 0,
+      selectedIcon: '★',
+      deselectedIcon: '☆'
+    };
+
+    let outOf = props.outOf ? props.outOf : 5;
```
JuliaShavyrova: expression can be simplified

src/components/StarRating/StarRating.jsx
```diff
@@ -0,0 +1,59 @@
...
+class StarRating extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      stars: [],
+      rating: 0,
+      hovered: 0,
+      selectedIcon: '★',
+      deselectedIcon: '☆'
```
JuliaShavyrova: why do you need to store icons in state?

src/components/StarRating/StarRating.jsx
```diff
@@ -0,0 +1,59 @@
...
+      <S.RateWrapper>
+        <S.StarWrapper>
+          {stars.map(star => (
+            <S.Star
+              key={star}
+              onClick={() => this.changeRating(star)}
+              onMouseEnter={() => this.hoverRating(star)}
+              onMouseLeave={() => this.hoverRating(0)}
+            >
+              {rating < star ? (hovered < star ? deselectedIcon : selectedIcon) : selectedIcon}
```
JuliaShavyrova: expression can be simplified

src/components/StarRating/StarRating.jsx
```diff
@@ -0,0 +1,59 @@
...
+      rating: 0,
+      hovered: 0,
+      selectedIcon: '★',
+      deselectedIcon: '☆'
+    };
+
+    let outOf = props.outOf ? props.outOf : 5;
+
+    for (let i = 0; i < outOf; i++) {
+      this.state.stars.push(i + 1);
```
JuliaShavyrova: hint: you can try to use es6 array methods for generate array a bit shorter way

.gitignore
GedasIncirauskas: Ok, fixed.

src/components/StarRating/StarRating.jsx
GedasIncirauskas: Fixed.

src/components/StarRating/StarRating.jsx
GedasIncirauskas: Fixed, I created separate file for icons.

src/components/StarRating/StarRating.jsx
GedasIncirauskas: Fixed, wrote a simplified expression.

src/components/StarRating/StarRating.jsx
GedasIncirauskas: Thank you, I have changed.

src/components/Form/Form.jsx
GedasIncirauskas: Ok.

src/components/StarRating/StarRating.jsx
```diff
@@ -0,0 +1,54 @@
...
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      stars: [],
+      rating: 0,
+      hovered: 0
+    };
+
+    this.state.stars.push(...Array.from({ length: 5 }, (_, i) => i + 1));
```
JuliaShavyrova: please move `...Array.from({ length: 5 }, (_, i) => i + 1)` to `stars` property definition, no need to use `.push()` for that

GedasIncirauskas: Ok, thanks. Already fixed.

PR #208, opened 2022-05-12
created controlled component

GedasIncirauskas:
1. Created controlled star component.

### DIFF
--- a/src/components/Form/Form.jsx
+++ b/src/components/Form/Form.jsx
@@ -12,8 +12,11 @@ const isFeedbackEmpty = value => value.length;
 
 const BasicForm = () => {
   const [message, setMessage] = useState('');
-  const [rating, setRating] = useState(null);
   const [formIsValid, setFormIsValid] = useState(false);
+  const [starReview, setStarReview] = useState({
+    stars: [...Array.from({ length: 5 }, (_, i) => i + 1)],
+    rating: 0
+  });
 
   const {
     value: userName,
@@ -52,7 +55,7 @@ const BasicForm = () => {
       return;
     }
 
-    Storage.setReview({ userName, emailValue, reviewValue, rating });
+    Storage.setReview({ userName, emailValue, reviewValue, rating: starReview.rating });
 
     resetName();
     resetEmail();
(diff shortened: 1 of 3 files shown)


src/components/Form/Form.jsx
```diff
@@ -12,8 +12,11 @@ const isFeedbackEmpty = value => value.length;
 
 const BasicForm = () => {
   const [message, setMessage] = useState('');
-  const [rating, setRating] = useState(null);
   const [formIsValid, setFormIsValid] = useState(false);
+  const [starReview, setStarReview] = useState({
+    stars: [...Array.from({ length: 5 }, (_, i) => i + 1)],
```
JuliaShavyrova: if `stars` is a variable that never changes, it would be better to store it as constant, not in state object

src/components/StarRating/StarRating.jsx
```diff
@@ -7,40 +7,32 @@ class StarRating extends Component {
...
     this.props.onChange(star);
   }
 
+  hoverRating(star) {
+    this.setState({ hoverAnimation: star });
+  }
+
   render() {
-    const { stars, rating, hovered } = this.state;
+    const { stars, rating } = this.props.value;
```
JuliaShavyrova: no need to pass an object in `value` prop, make it simplier :)

src/components/Form/Form.jsx
GedasIncirauskas: Ok, I will fix it ;) pass it to constants.

src/components/StarRating/StarRating.jsx
GedasIncirauskas: Hope fixed in the right way :)))

src/components/Form/Form.jsx
```diff
@@ -12,8 +12,10 @@ const isFeedbackEmpty = value => value.length;
 
 const BasicForm = () => {
   const [message, setMessage] = useState('');
-  const [rating, setRating] = useState(null);
   const [formIsValid, setFormIsValid] = useState(false);
+  const [starReview, setStarReview] = useState({ rating: 0 });
```
JuliaShavyrova: why do you store object in this state?

GedasIncirauskas: Thanks, removed value as object.

PR #212, opened 2022-05-20
fixed webpack, created 404 page, updated routes

GedasIncirauskas:
1. Fixed `Cannot GET / error` with webpack.
2. Created 404 page for bad URL's.
3. Updates routes and page tabs title.

Annotation 2022-05-20 145207


src/config/Routes/Routes.jsx
```diff
@@ -0,0 +1,56 @@
...
+  },
+
+  {
+    path: '/*',
+    Component: NotFound,
+    title: translations.msg_page_not_found_404_error
+  }
+];
+
+const Routes = () => {
```
JuliaShavyrova: it's a component, not a config, please move it in more convenient place

src/App.jsx
```diff
@@ -10,19 +9,14 @@ const App = () => {
   const { isDarkMode } = useContext(DarkModeContext);
 
   return (
-    <HashRouter>
+    <>
```
JuliaShavyrova: why no other router defined instead of HashRouter?

src/config/Routes/Routes.jsx
```diff
@@ -0,0 +1,56 @@
...
+  return (
+    <Switch>
+      {routes.map(({ path, Component, isExact, title }) => (
+        <Route
+          key={path}
+          path={path}
+          exact={isExact}
+          render={() => {
+            document.title = title;
```
JuliaShavyrova: good idea, interesting implementation :)
the only thing i'd like to note - "page not found" looks for me not very informative just because i don't understand what exact app shows it (without opening this tab). so i'd suggest to add favicon to you website and everything would look great

src/App.jsx
GedasIncirauskas: I used `{ BrowserRouter as Router }` at my index.js file, so I don't need here.
For better understanding I will pass it in the same file as `Routes`

src/config/Routes/Routes.jsx
GedasIncirauskas: Ok, fixed.

src/config/Routes/Routes.jsx
GedasIncirauskas: Ok, thanks ;) 
I will add more information.

image

JuliaShavyrova (approved):

PR #216, opened 2022-05-24
Redux implementation

mrvanagas:
Implemented Redux. Now SearchBar and MultiCard components can be renderred independently of each other, allowing for those components to be potentially reused in other parts of the app. 

Current issue - when sending a second API request for a different City, the cards duplicate, adding 7 more cards of the other city forecast:
image
Investigating why it is like that. Potentially this can lead to a new, modified feature.


client/src/App.jsx
```diff
@@ -12,10 +12,10 @@ function App() {
...
         <ul>
           <li>
-            <Link to="/">Home</Link>
-            <Link to="/about">About</Link>
-            <Link to="/contact">Contact</Link>
-            <Link to="/cities">Search</Link>
+            <Link to="/"> Home</Link>
+            <Link to="/about"> About</Link>
+            <Link to="/contact"> Contact</Link>
+            <Link to="/cities"> Search</Link>
```
JuliaShavyrova, line 18: if you need spaces near every link, it would be better to add styling

client/src/Components/ContactForm.jsx
```diff
@@ -0,0 +1,94 @@
+import React from 'react';
```
JuliaShavyrova, line 1: i've already seen this file in previous PR, please check branch and PR again to make sure it contains actual changes only

client/src/App.jsx
mrvanagas: This will be styled once Header and Footer are added.

client/src/Components/ContactForm.jsx
mrvanagas: I am not quite sure for this behavior, tho from what I have seen in the PR, there are changes adding redux in this branch.

PR #223, opened 2022-05-30
add first simple test, change naming for Table component

GedasIncirauskas:
1. Change component title `Table` to `ForecastTable`.
2. Installed `@testing-library/jest-dom`  and `@testing-library/react`.
3. Written the first test for `Form` component.

Screenshot 2022-05-30 195353


package.json
```diff
@@ -19,6 +19,7 @@
     "npm": ">=3"
   },
   "dependencies": {
+    "@testing-library/react": "^12.1.2",
```
JuliaShavyrova: please move to dev dependencies

src/components/Form/Form.test.js
```diff
@@ -0,0 +1,19 @@
+import { render, screen } from '@testing-library/react';
+import Form from './Form';
+import { translations } from '../../utils/translations';
+import '@testing-library/jest-dom/extend-expect';
```
JuliaShavyrova: why do you need this import?

src/components/Form/Form.test.js
```diff
@@ -0,0 +1,19 @@
...
+    const inputName = screen.getByText(translations.msg_user_name_label);
+    const inputEmail = screen.getByText(translations.msg_user_email_label);
+    const textareaFeedback = screen.getByText(translations.msg_user_review_label);
+    const submitButton = screen.getByRole('button');
+
+    expect(inputName).toBeInTheDocument();
+    expect(inputEmail).toBeInTheDocument();
+    expect(textareaFeedback).toBeInTheDocument();
+    expect(submitButton).toBeInTheDocument();
+  });
```
JuliaShavyrova, line 35: please add more tests to check form functionality (at least validations, submitting)

src/pages/NotFoundPage/NotFoundPage.jsx
```diff
@@ -1,9 +1,9 @@
 import { Link } from 'react-router-dom';
 import { translations } from '../../utils/translations';
-import pageNotFoundIcon from '../../utils/icons/notfound.png';
 import * as S from './NotFoundPage.styles';
 
 const NotFoundPage = () => {
+  const pageNotFoundIcon = require('../../utils/icons/notfound.png');
```
JuliaShavyrova: why do you use `require` here?

GedasIncirauskas: Because if I importing it like es6 syntax I have error in my test. I tried change webpack setting, but unfortunately only in this way it works without any errors.. :) 

Error 2022-05-30 195703

package.json
GedasIncirauskas: Ok, fixed

src/components/Form/Form.test.js
GedasIncirauskas: Because I have error:
image


So to fit it I need this:
image

src/components/Form/Form.test.js
GedasIncirauskas: Ok, I will add more.

src/pages/NotFoundPage/NotFoundPage.jsx
JuliaShavyrova: you need to update your app build configuration to make it importing png files correctly

src/components/Form/Form.test.js
JuliaShavyrova: usually it's enough to import jest-dom once in some file like src/setupTests.js and define this setup in [config](https://jestjs.io/docs/configuration#setupfiles-array) if needed

GedasIncirauskas: Thank you for advice. Fixed and works fine 👍

src/pages/NotFoundPage/NotFoundPage.jsx
GedasIncirauskas: Error fixed.

JuliaShavyrova (approved):

PR #224, opened 2022-06-03
refactoring code in some places

GedasIncirauskas:
1. Refactoring code (written simpler code, removed unused elements).

### DIFF
--- a/src/components/CurrentWeather/CurrentWeather.jsx
+++ b/src/components/CurrentWeather/CurrentWeather.jsx
@@ -8,7 +8,7 @@ import * as S from './CurrentWeather.styles';
 const CurrentWeather = ({ data }) => {
   const [weatherData] = data;
 
-  if (data.length === 0) return <Spinner />;
+  if (!weatherData) return <Spinner />;
 
   return (
     <S.Container>
--- a/src/components/CurrentWeather/CurrentWeather.jsx
+++ b/src/components/CurrentWeather/CurrentWeather.jsx
@@ -8,41 +8,37 @@ import * as S from './CurrentWeather.styles';
 const CurrentWeather = ({ data }) => {
   const [weatherData] = data;
 
+  if (data.length === 0) return <Spinner />;
+
   return (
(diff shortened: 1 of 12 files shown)


src/components/CurrentWeather/CurrentWeather.jsx
```diff
@@ -8,41 +8,37 @@ import * as S from './CurrentWeather.styles';
 const CurrentWeather = ({ data }) => {
   const [weatherData] = data;
 
+  if (data.length === 0) return <Spinner />;
```
JuliaShavyrova: expression can be simplified

src/pages/HourlyForecast/HourlyForecast.jsx
```diff
@@ -24,26 +24,24 @@ const HourlyForecast = () => {
...
-            />
-          ))}
-        </>
-      ) : (
-        <Spinner />
-      )}
-    </div>
+    <>
+      <h2>{`${translations.msg_page_hourly_forecast_h1} ${city}`}</h2>
+      {hourlyForecast?.map(({ symbol, time, temperature, windSpeed }, index) => (
```
JuliaShavyrova: i can suggest to use ternary operator to avoid unnecessary `?`. for example, here (and in some other places) you can refactor code like this:
`<h2>...</h2>`
`   {!hourlyForecast ? <Spinner/> : hourlyForecast.map(...)}`

also there is no need to hide title while page is loading, because it's hardcoded (no requests data used for it) and helps user additionally to understand what page is opened

GedasIncirauskas: Ok, thank you for advice. I will fix it.

src/components/CurrentWeather/CurrentWeather.jsx
GedasIncirauskas: Ok, thank you. I fixed 👍

JuliaShavyrova (approved):

PR #241, opened 2022-06-27
Types added

ziviledauderiene:


src/helpers/api.ts
```diff
@@ -1,6 +1,7 @@
-import axios from 'axios';
+import axios, { AxiosResponse } from 'axios';
+import { AxiosOptions, Url } from './helpersInterfaces';
 
-export const URL = {
+export const URL: Url = {
```
JuliaShavyrova: `Url` interface (and variable) name is a bit confusing, please rename it to define that it's a config for api routes

src/helpers/helpersInterfaces.ts
```diff
@@ -0,0 +1,20 @@
...
+  observations: string;
+  current: string;
+  nowCast: string;
+  hourly: string;
+  threeHourly: string;
+  daily: string;
+  airQuality: string;
+}
+
+export interface AxiosOptions {
```
JuliaShavyrova: you can use AxiosRequestConfig built-in interface to define required options for your requests

ziviledauderiene: ok, changed. thank you

src/helpers/api.ts
```diff
@@ -1,6 +1,7 @@
-import axios from 'axios';
+import axios, { AxiosRequestConfig, AxiosResponse } from 'axios';
+import { UrlInterface } from './helpersInterfaces';
 
-export const URL = {
+export const URL: UrlInterface = {
```
JuliaShavyrova: 1. your const URL is definitely not a url, it's a config for api routes you going to use in app, so please rename, the same for interface - it's not describing a url model, it describes very different stuff
2. no need to use word `Interface/Type` when you're creating an interface/type. One of the reasons: if you'll need to transform it from interface to type in future, you'll need to rename and replace it everywhere. You can follow [this](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines) coding guideline as example

JuliaShavyrova (approved):

PR #247, opened 2022-07-11
Feedback Form component

Lvlad27:
- Added Feedback Form component
- Webpack index.html source file change
- dataService file restructure

image


src/Components/FeedbackPage/FeedbackForm/FeedbackForm.jsx
```diff
@@ -0,0 +1,204 @@
...
+        ...prevFormData,
+        [name]: type === 'checkbox' ? checked : value
+      };
+    });
+  }
+
+  function handleSubmit(event) {
+    event.preventDefault();
+    setFormErrors(validate(formData));
+    setIsSubmit(true);
```
JuliaShavyrova: why not just save to local storage when `validate(formData)` returns empty object? in this case you wouldn't need isSubmit state at all

src/Components/FeedbackPage/FeedbackForm/FeedbackForm.jsx
```diff
@@ -0,0 +1,204 @@
...
+    if (!values.comments) {
+      errors.comments = '❌ Please leave a comment!';
+    } else if (!values.comments.length < 20) {
+      errors.comments = '❌ Must be at least 20 characters';
+    }
+    if (!values.feature) {
+      errors.feature = '❌ Please select a feature!';
+    }
+    if (values.feedbackType === 'selection' || !values.feedbackType) {
+      errors.feedbackType = '❌ Please select feedback type!';
```
JuliaShavyrova: i would recommend to move the cross icon into ErrorMessage component as far as it's not dynamic and should be shown for any error

src/Components/FeedbackPage/FeedbackForm/FeedbackForm.jsx
```diff
@@ -0,0 +1,204 @@
...
+          <input
+            type="radio"
+            id="forecast"
+            name="feature"
+            value="Forecast"
+            checked={formData.feature === 'Forecast'}
+            onChange={handleChange}
+          />
+          <label htmlFor="forecast">Forecast</label>
+          <br />
```
JuliaShavyrova: br tag is usually used for text formatting (inside `<p></p>` or `<span></span>` tags) not blocks, so would be better to use css instead in this case

src/Components/FeedbackPage/FeedbackForm/FeedbackForm.jsx
Lvlad27: I agree. Removed isSubmit state.

src/Components/FeedbackPage/FeedbackForm/FeedbackForm.jsx
Lvlad27: done

src/Components/FeedbackPage/FeedbackForm/FeedbackForm.jsx
Lvlad27: done

JuliaShavyrova (approved):

PR #248, opened 2022-07-13
Dark/light theme

ziviledauderiene:
https://user-images.githubusercontent.com/96768501/178669108-3c4c8c25-a071-4574-a3a7-11746293eb24.mp4

- dark/light theme switcher implemented using React.createContext();
- minor change in server.js: token expiring time added.


src/store/theme-context.tsx
```diff
@@ -0,0 +1,45 @@
+import React, { useEffect, useState } from 'react';
+
+export interface ThemeContextConfig {
+  theme: string;
+  toggleTheme?: () => void;
+}
+
+const themes: {
```
JuliaShavyrova: as far as you are using typescript you could just create enum for theme, it both helps to add types (at least for useState in your case) can be used as regular variable.
also i'd recommend to replace empty string for `light` property to "light"

ziviledauderiene: Ok, changed to enum.

JuliaShavyrova (approved):

PR #252, opened 2022-07-21
Favourite Cities feature markup and functionality

Lvlad27:
- user can now add any city as a favourite and will be added to the list of favoured cities, which can be viewed in Favourites page. User can also remove city from favorite.
     -  Later maybe I'll add more functionality to this such as deleting or searching and favouring a city directly from Favourites Page as well.
- implemented redux and persisted state in local storage

[Weather-App.webm](https://user-images.githubusercontent.com/86689859/180215302-d83b875d-67a9-400b-9013-067aaea3ea37.webm)


src/Components/FavPage/FavPage.jsx
```diff
@@ -1,9 +1,27 @@
...
+    return <FavCitiesList data={item} key={item.locationData.id} />;
+  });
+
   return (
     <main className={styles.container}>
-      <h2>Favourite Locations</h2>
+      {listItems.length !== 0 && <h2 className={styles.title}>Favourite Locations</h2>}
+
+      {listItems.length !== 0 ? (
+        <div className={styles['container-items']}>{listItems}</div>
```
JuliaShavyrova: why not combine two conditions in one like this?
if (listItems.length !== 0) {
  < title >
  < items >
} else {
  < message > 
}

Lvlad27: Yes, that seems better. Too much reading about conditional rendering :) . Changes made.

src/Components/FavPage/FavPage.jsx
```diff
@@ -1,11 +1,34 @@
...
+  const favCityList = useSelector((state) => state);
+
+  const listItems = favCityList.map((item) => {
+    return <FavCitiesList data={item} key={item.locationData.id} />;
+  });
+
+  function renderComponent() {
+    if (listItems.length !== 0) {
+      return (
+        <main className={styles.container}>
```
JuliaShavyrova: you could just use react fragment instead of moving `main` tag :)
also no need to create separate `renderComponent` function

src/Components/FavPage/FavPage.jsx
```diff
@@ -1,11 +1,34 @@
...
+          <h2 className={styles.title}>Favourite Locations</h2>
+          <div className={styles['container-items']}>{listItems}</div>
+        </main>
+      );
+    } else {
+      return (
+        <Message
+          message1={'Here you can see a list of the weather from your favourite cities.'}
```
JuliaShavyrova: forgot to add before few suggestions about `Message` component: 
1.`message1` and `message2` prop names are not very clear, what's the difference between them?
2. why not use one `Message` component twice?
<Message>{'Here you can see a list of the weather from your favourite cities.'}</Message>
<Message>{'Your list is empty. Search for a city and add it as a favourite!'}</Message>
3. or why not use `children` prop for content and pass any jsx inside `Message` component?

src/Components/FavPage/FavPage.jsx
Lvlad27: Ok to use react fragment, but if I don't create renderComponent function how can I use if statement in the return statement of the FavPage Component?  Also can't use ternary because I have multiline statements in the first condition (h2 + div).

This does not work:

image

JuliaShavyrova: just use the same ternary operator you used before
listItems.length !== 0 ? (<>...</>) : (<>...</>)
difference is that previously you had two different conditions (first - listItems.length !== 0 && ..., second - listItems.length !== 0 ? (<>...</>) : (<>...</>)), but now only one

src/Components/FavPage/FavPage.jsx
Lvlad27: Ok, thank you for suggestions, I've done the changes.

JuliaShavyrova (approved):

PR #253, opened 2022-07-22
Current location and Previous searches added to the dropdown list as suggestions

ziviledauderiene:
screenshot030

- index-redux.ts created, which combined two reducers (tempUnits and location);
- Current location and Previous searches added to the dropdown list as suggestions. It appears immediately on input focus. When something is typed to the input, current and prev are shown in the end.
- 5 prev searches are saved as global variables and in local storage to persist reloads. If a user searches same location again, it is not added for the second time, but moved to the top (as the last search) in the list.


src/components/Home/Search/SearchResults/SearchResults.tsx
```diff
@@ -18,23 +29,52 @@ const SearchResults: React.FunctionComponent<SearchResultsProps> = ({
...
   const { setLocationId } = useContext(LocationContext);
+  const { currentLocation, prevSearches } = useSelector<CombinedState, LocationState>(
+    state => state.location
+  );
+  const dispatch: Dispatch<LocationActionConfig> = useDispatch();
+
+  const clickHandler = useCallback(
+    (event: React.MouseEvent<HTMLDivElement, MouseEvent>) => {
+      const target = event.target as HTMLDivElement;
+      console.log('clickHandler', target.id);
```
JuliaShavyrova: please don't forget to clean your code :)

src/store/location-redux.ts
```diff
@@ -0,0 +1,71 @@
...
+export interface LocationActionConfig {
+  type: LocationActions;
+  payload: string;
+}
+
+const getPrevSearches = (): string[] => {
+  const prevSearches: string | null = localStorage.getItem('prevSearches');
+  if (prevSearches) {
+    return JSON.parse(localStorage.getItem('prevSearches')!);
+  } else return [];
```
JuliaShavyrova: could be used ternary operator instead. or at least please use the same syntax for if and else conditions

ziviledauderiene: ok, changed to ternary operator

JuliaShavyrova (approved):

PR #260, opened 2022-07-29
Fe light dark temperature controls

Kolyo-Coherent:


src/App.jsx
```diff
@@ -20,19 +23,43 @@ export default function App() {
...
+    if (!theme) {
+      if (!!cashedTheme) {
+        themeElement.dataset.theme = cashedTheme;
+        setTheme(cashedTheme);
+      } else {
+        themeElement.dataset.theme = 'light';
+      }
+    } else {
+      sessionStorage.setItem('theme', theme);
+      document.getElementById('app').dataset.theme = theme;
```
JuliaShavyrova: 1. why not `themeElement.dataset.theme = theme;` instead of `document.getElementById('app').dataset.theme = theme;`?
2. as far as i see, in every condition you assign new theme for `themeElement`, so line with assignment can be moved out of condition blocks and new theme value can be defined like this:
`const newTheme = theme || cashedTheme || "light";`

JuliaShavyrova (approved):

PR #262, opened 2022-08-02
Fe forma and about

Kolyo-Coherent:


src/components/FeedbackForm.jsx
```diff
@@ -0,0 +1,106 @@
...
+          .catch(err => {
+            console.error('error occured: ', err.message);
+          });
+      }, 5000);
+    }
+  };
+
+  return (
+    <div>
+      {valid ? <div className="success-msg">Form submitted successfully!</div> : null}
```
JuliaShavyrova:
```suggestion
      {valid && <div className="success-msg">Form submitted successfully!</div>}
```

src/components/FeedbackForm.jsx
```diff
@@ -0,0 +1,106 @@
...
+        <fieldset>
+          <label htmlFor="firstName">First name:</label>
+          <input
+            type="text"
+            id="firstName"
+            name="firstName"
+            value={formValues.firstName}
+            onChange={changeVal}
+          />
+          {submitted && !formValues.firstName ? (
```
JuliaShavyrova: the same as suggestion above (for other messages too)

src/components/FeedbackForm.jsx
```diff
@@ -0,0 +1,106 @@
...
+  };
+
+  const formSubmit = e => {
+    e.preventDefault();
+    setSubmitted(true);
+
+    if (formValues.firstName && formValues.lastName && formValues.mail && formValues.feedback) {
+      setValidity(true);
+      setTimeout(async function () {
+        await axios
```
JuliaShavyrova: please check requirements and save data in local storage

src/components/FeedbackForm.jsx
```diff
@@ -0,0 +1,106 @@
...
+    } else {
+      alert('Something went wrong! Please try again.');
+    }
+  };
+
+  const formSubmit = e => {
+    e.preventDefault();
+    setSubmitted(true);
+
+    if (formValues.firstName && formValues.lastName && formValues.mail && formValues.feedback) {
```
JuliaShavyrova: you can destructure `formValues` object here

Kolyo-Coherent: how?

JuliaShavyrova:
```
const { firstName, .... } = formValues;
if (firstName && ....) {
```

src/components/FeedbackForm.jsx
```diff
@@ -0,0 +1,185 @@
...
+  const formSubmit = e => {
+    e.preventDefault();
+    setSubmitted(true);
+    const { firstName, lastName, mail, rating, feedback } = formValues;
+    if (firstName && lastName && mail && rating && feedback) {
+      setValidity(true);
+      setTimeout(function () {
+        localStorage.setItem('survey-form', JSON.stringify(formValues));
+        resetForm({ status: 200 });
+      }, 5000);
```
JuliaShavyrova, line 58: why do you need timeout while saving in local storage?

src/components/FeedbackForm.jsx
```diff
@@ -0,0 +1,185 @@
...
+            </div>
+            <div className="radio-set">
+              <input
+                type="radio"
+                id="satisfactoryIndex3"
+                name="rating"
+                value="mediocre"
+                onChange={radioSelected}
+              />
+              <label htmlFor="satisfactoryIndex3">&#128547;</label>
```
JuliaShavyrova: i would recommend to create config array for rating like this and then map inside jsx

```
const ratingConfig = [
  { value: 'super-bad', label: '&#128169;' }
  ....
];
...
ratingConfig.map(({ value, label }, index) => (
  <div className="radio-set">
  .......
```

src/components/FeedbackForm.jsx
Kolyo-Coherent: to see the success message before reset

src/components/FeedbackForm.jsx
Kolyo-Coherent: ok

src/components/FeedbackForm.jsx
JuliaShavyrova: please rewrite your code to not use setTimeout, it's redundant

Kolyo-Coherent: any suggestions how to do it?

JuliaShavyrova: check again logic of showing success message and update it. 
usually success message can be shown after form successfully submitted and disappear after some timeout (but it definitely should not depend on saving form process, in your case form should be saved immediately) or it can be left until user clicks submit button again

src/components/FeedbackForm.jsx
```diff
@@ -0,0 +1,140 @@
...
+
+  const radioOptions = [
+    { val: "super-bad", emoji: '\u{1F4A9}' },
+    { val: "bad", emoji: '\u{1F922}' },
+    { val: "mediocre", emoji: '\u{1F623}' },
+    { val: "OK", emoji: '\u{1F914}' },
+    { val: "good", emoji: '\u{1F60A}' },
+    { val: "very-good", emoji: '\u{1F60D}' },
+    { val: "excellent", emoji: '\u{1F37A}' },
+  ]
```
JuliaShavyrova: please move variable outside of the component body to prevent its re initiating every time the component re-renders

src/components/FeedbackForm.jsx
```diff
@@ -0,0 +1,140 @@
...
+      resetForm({ status: 200 });
+      setTimeout(function () {
+        setValidity(false);
+      }, 5000);
+    }
+  };
+
+  return (
+    <div>
+      {valid && <div className="success-msg">Form submitted successfully!</div>}
```
JuliaShavyrova: i'm a bit confused that "valid" variable is responsible of showing message after form submitted and "submitted" variable is used for showing validation messages. don't you think it would be better to swap them? :)

proxy.js
```diff
@@ -116,4 +116,9 @@ app.post('/search', (req, res) => {
     .catch(error => console.log('error', error));
 });
 
+app.post('/send-feedback', (req, res) => {
+  const formFields = req.body;
+  res.send('Form submitted successfully!');
+});
```
JuliaShavyrova: please remove if it's not used anywhere

src/components/FeedbackForm.jsx
```diff
@@ -0,0 +1,140 @@
...
+          {submitted && !formValues.mail && (
+            <span className="error-msg">E-mail is empty or invalid!</span>
+          )}
+        </fieldset>
+        <fieldset>
+          <label htmlFor="satisfactory-index">Rate our services</label>
+          <div className="radio-set-container">
+            {radioOptions.map((obj, idx) => {
+              const a = new DOMParser().parseFromString(obj.emoji, "text/xml")
+              console.log(a)
```
JuliaShavyrova: reminding again about cleaning code

JuliaShavyrova (approved):

PR #263, opened 2022-08-02
Responsive design, bug fixes, About component

ziviledauderiene:
Fixes/changes/improvements according to these comments: 

- contacts page: please add validation error when filling "number" input with letters, because now it shows no error while submit button is disabled. also after form successfully submitted inputs are cleared but submit button is not disabled
- about page: just add some content :)
- also would be great to have adaptivity
- home page: found one more bug when selecting current location (video attached) + also noticed url like http://localhost:9020/cities/25.2797933,54.7062492 that is broken after page reload


src/components/Header/Header.tsx
```diff
@@ -1,60 +1,47 @@
...
   const { theme, toggleTheme }: ThemeContextConfig = useContext(ThemeContext);
   const dispatch: Dispatch<TempActionConfig> = useDispatch();
   const tempUnit = useSelector<CombinedState, TempUnits>(state => state.tempUnit);
   const isCelsius: boolean = tempUnit === TempUnits.CELSIUS;
+  const [showMobileNav, setShowMobileNav] = useState<boolean>(false);
 
   const toggleTempUnits = (event: React.ChangeEvent<HTMLSelectElement>) => {
     dispatch({ type: event.target.value });
   };
+  const burgerClickHAndler = (): void => {
```
JuliaShavyrova: just a typo
```suggestion
  const burgerClickHandler = (): void => {
```

src/components/Home/Today1Hourly/Today1Hourly.tsx
```diff
@@ -53,6 +54,9 @@ class Today1Hourly extends React.Component<Today1HourlyProps, Today1HourlyState>
 
   render() {
     const { data, loading, error } = this.state;
+    if (window.screen.width < 650) {
```
JuliaShavyrova: i would recommend to move the condition in parent component. reason: logic of showing/hiding component doesn't belong to component itself and you might have another page where you need to show it but with no restrictions about screen width

ziviledauderiene: ok, moved

JuliaShavyrova (approved):

PR #264, opened 2022-08-03
Testing theme and location reducers

Lvlad27:


src/__tests__/locations.test.js
```diff
@@ -0,0 +1,10 @@
+import cityReducer from '../redux/locations';
+
+describe('theme reducer', () => {
+  it('should return the initial state when passed an empty action', () => {
+    const favCityList = [];
+    const action = { type: '' };
+    const result = cityReducer(favCityList, action);
+    expect(result).toEqual([]);
+  });
```
JuliaShavyrova, line 10: looks good, please add two more test cases (with different actions provided)

package.json
```diff
@@ -83,6 +93,7 @@
     "pretty-quick": "1.11.1",
     "react-hot-loader": "4.13.0",
     "redux-immutable-state-invariant": "2.1.0",
+    "regenerator-runtime": "^0.13.9",
```
JuliaShavyrova: why do you need this package?

Lvlad27: I was getting similar error from this post and just followed the answer to try and fix the error:
[https://stackoverflow.com/questions/69091261/syntaxerror-support-for-the-experimental-syntax-decorators-legacy-isnt-curre](url). But it didn't work so I will remove it.

src/__tests__/locations.test.js
```diff
@@ -0,0 +1,25 @@
...
+    const favCityList = [];
+    const cityName = 'Bucharest';
+    const action = { type: 'SAVE_FAV_CITY', payload: cityName };
+    const result = cityReducer(favCityList, action);
+    expect(result).toEqual(['Bucharest']);
+  });
+  it('should delete city data object', () => {
+    const favCityList = mockCityData;
+    const cityName = 'Bucharest';
+    const action = { type: 'DELETE_FAV_CITY', payload: cityName };
```
JuliaShavyrova: did this test pass?
as far as i understand, your reducer removes location comparing ids, not city names

Lvlad27: Yes that's correct, but action.payload.locationData.id is the id of the given city name, action.payload being the city name. It compares each of the array's items id with the id of the given city name.

image
image

Lvlad27: fix test case and added extra one

image

JuliaShavyrova (approved):

PR #266, opened 2022-08-04
Refactor & fixes

ziviledauderiene:
- param && accessToken check added to prevent bad requests (400 and 401);
- styling fixes;
- compressed images;
- one axios import left;
- grouped imports.


src/components/Home/Search/SearchResults/SearchResults.tsx
```diff
@@ -50,7 +52,7 @@ const SearchResults: React.FunctionComponent<SearchResultsProps> = ({
     },
     [event]
```
JuliaShavyrova: why do you have event dependency here? and please don't forget to update dependencies arrays

src/components/Home/Search/Search.tsx
```diff
@@ -39,14 +38,13 @@ const Search: React.FunctionComponent = () => {
...
-          setSearchTerm={setSearchTerm}
-          setDisplaySearchResults={setDisplaySearchResults}
-        />
-      )}
+      <SearchResults
+        searchResults={searchResults}
+        searchLoading={searchLoading}
+        setSearchTerm={setSearchTerm}
+        setDisplaySearchResults={setDisplaySearchResults}
+        displaySearchResults={displaySearchResults}
```
JuliaShavyrova, line 35: seems like there are too many props provided :)
what can be improved: 
1. you can move location search request inside search results, so `searchResults ` and `searchLoading` won't be needed at all, but you'll need to app `searchTerm` prop instead
2. `setSearchTerm` and `setDisplaySearchResults` also can be removed and replaced with callback prop like `onResultSelected/resultSelectedHandler`, and in SearchResults component you can just call onResultSelected() instead of 
```
setSearchTerm('');
setDisplaySearchResults(false);
```
it will allow you to keep all logic about these two states in one place

src/components/Home/Search/SearchResults/SearchResults.tsx
ziviledauderiene: must be [currentLocation]?

JuliaShavyrova: not only `currentLocation`, `navigate` and `dispatch` also will be a dependencies. 
please use 'eslint-plugin-react-hooks' with "exhaustive-deps" rule to define missed dependencies automatically

JuliaShavyrova: btw `setSearchTerm` and `setDisplaySearchResults` are also dependencies :)

JuliaShavyrova (approved):

PR #270, opened 2022-08-23
Add layout

AndrewHelix:


src/styles/normalizez.css
```diff
@@ -0,0 +1,344 @@
+/*! normalize.css v8.0.1 | MIT License | github.com/necolas/normalize.css */
```
JuliaShavyrova, line 1: small suggestion (not required to do): you can install this file using npm to not store in your src folder files that won't be updated ever.

JuliaShavyrova (approved):

PR #272, opened 2022-08-24
initial: routes, eslint, typescript configuration

Auryte:


src/App.tsx
```diff
@@ -0,0 +1,19 @@
+// @ts-ignore
+import Routes from 'components/Routes.tsx';
+// @ts-ignore
+import React from 'react';
+// @ts-ignore
```
JuliaShavyrova: ts-ignore should be used for some exceptional cases only, seems like the app not fully configured for using typescript

tsconfig.json
```diff
@@ -0,0 +1,24 @@
...
+    "paths": {
+      "*": ["src/*"]
+    },
+    "lib": ["esnext", "dom", "dom.iterable"],
+    "sourceMap": true,
+    "noImplicitAny": false,
+    "plugins": [{ "name": "typescript-plugin-css-modules" }]
+  },
+  "exclude": ["node_modules", "build"],
+  "include": ["./src", "webpack.config.js", "types"]
```
JuliaShavyrova: why do you need to include webpack.config.js here?

tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
+  "compilerOptions": {
+    "module": "commonjs",
```
JuliaShavyrova: why do you have "commonjs" value? your src files uses es modules

types/css.d.ts
```diff
@@ -0,0 +1,13 @@
+declare module '*.scss' {
+  const css: { [key: string]: string };
+  export default css;
+}
+declare module '*.sass' {
+  const css: { [key: string]: string };
+  export default css;
+}
+declare module 'react-markup';
```
JuliaShavyrova, line 9: why this type is declared? are you using some external module with the same name?

webpack/webpack.prod.js
```diff
@@ -55,6 +55,8 @@ module.exports = {
           MiniCssExtractPlugin.loader,
           {
             loader: 'css-loader',
+            // loader: 'babel-loader',
```
JuliaShavyrova: please remove commented code here

PR #276, opened 2022-08-26
Add router and find position

AndrewHelix:


src/components/app/App.tsx
```diff
@@ -1,18 +1,34 @@
...
+import { Routes, Route, BrowserRouter } from 'react-router-dom';
+
+import { Layout } from 'components/Layout';
+import { Countries } from 'components/countries/Countries';
+import { Details } from 'components/details/Details';
+import { Feedback } from 'components/feedback/Feedback';
+import { Page404 } from 'components/page404/Page404';
 
 const App: FC = () => {
+  const { position, error } = usePosition();
```
JuliaShavyrova: why not use hook inside Main component?

AndrewHelix: I thought that I will use info about position in another components and the first time use props. Do it form App components more comfortable than inside components

JuliaShavyrova: in case you want to use the same data across the whole app would be better to use context, not hook + props, because it can cause props drilling like this:
App body (data about position here):
```
<>
 <Comp1 position={position} />
</>
```

Comp1 body (not using position data at all, but needs to get and pass it because Comp2 uses it):
```
<>
  <Comp2 position={position} />
</>
```

Comp2 body:
```
<>
 {*data about position needs to be used here*}
</>
```

JuliaShavyrova (approved):

PR #277, opened 2022-08-29
Added Header with child components, Redux and functionality for switching theme

ViliusGaizutis:


JuliaShavyrova:
please sync branches and resolve conflicts

src/components/Input/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+const INPUT = {
+  input: ({ type, placeholder, theme }: { type: string; placeholder: string; theme: string }) => (
+    <S.Input type={type} placeholder={placeholder} theme={theme} />
+  )
+};
+
+interface InputProps {
+  inputElement: 'input';
+  type: 'text' | 'textarea' | 'search';
```
JuliaShavyrova: input type cannot have "textarea" value; you can check all available types in mdn docs and use `HTMLInputTypeAttribute` built-in type for definition

src/components/Input/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+// Custom hooks
+import { useAppSelector } from '../../redux/hooks';
+
+// Styles
+import * as S from './styles';
+
+const INPUT = {
+  input: ({ type, placeholder, theme }: { type: string; placeholder: string; theme: string }) => (
+    <S.Input type={type} placeholder={placeholder} theme={theme} />
+  )
```
JuliaShavyrova: just interested - why do you need to create the separate const for input?

src/components/Input/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+};
+
+interface InputProps {
+  inputElement: 'input';
```
JuliaShavyrova, line 12: are you going to extend available set of values for 'inputElement' prop?

src/components/Input/index.tsx
ViliusGaizutis: My initial idea was to use switch statement, but as I learned from previous training that it is not advised, I decided to use declarative manner.

src/components/Input/index.tsx
ViliusGaizutis: Yes. Since there will be a feedback page with a form which will possibly include textarea elements. I haven't decided on this yet.

src/components/Input/index.tsx
ViliusGaizutis: My mistake, it should have been under "inputElement: 'input';".

src/components/Input/index.tsx
```diff
@@ -1,27 +1,30 @@
...
-  input: ({ type, placeholder, theme }: { type: string; placeholder: string; theme: string }) => (
-    <S.Input type={type} placeholder={placeholder} theme={theme} />
-  )
-};
+type InputType = HTMLInputTypeAttribute;
 
 interface InputProps {
   inputElement: 'input';
-  type: 'text' | 'textarea' | 'search';
+  type: InputType;
```
JuliaShavyrova, line 13:
```suggestion
  type: HTMLInputTypeAttribute;
```

JuliaShavyrova (approved):

PR #280, opened 2022-08-31
Added api service and search functionality

ViliusGaizutis:


src/types/index.ts
```diff
@@ -0,0 +1,17 @@
+export interface searchResultType {
```
JuliaShavyrova: please use PascalCase for interfaces names

src/common/components/Search/SearchResultList/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+import SearchResultItem from './SearchResultItem';
+
+// Styles
+import * as S from '../styles';
+
+// Types
+import { searchResultType } from '../../../../types';
+
+interface SearchResultListProps {
+  searchResults: [searchResultType];
```
JuliaShavyrova: if you meant array of SearchResultType, it should be written like this:
`searchResults: SearchResultType[];`

src/services/forecaApi.ts
```diff
@@ -0,0 +1,19 @@
+import { createApi, fetchBaseQuery } from '@reduxjs/toolkit/query/react';
+
+export const forecaApi = createApi({
+  reducerPath: 'forecaApi',
+  baseQuery: fetchBaseQuery({
+    baseUrl: 'https://pfa.foreca.com/api/v1/',
+    prepareHeaders: headers => {
+      headers.set('authorization', `Bearer ${process.env.API_KEY}`);
```
JuliaShavyrova: please implement silent login process, not just using token stored in .env file

src/common/components/Search/SearchResultList/SearchResultItem/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+// Types
+import { searchResultType } from '../../../../../types';
+
+interface SearchResultItemProps {
+  searchResult: searchResultType;
+}
+
+const SearchResultItem = ({ searchResult }: SearchResultItemProps) => {
+  const searchResultItemRef = useRef<HTMLLIElement>(null);
```
JuliaShavyrova: no need to use ref here

server/index.js
```diff
@@ -0,0 +1,39 @@
...
+    res
+      .cookie('token', token, {
+        expires: new Date(Date.now() + expiresIn * 1000),
+        httpOnly: true,
+        secure: req.secure || req.headers['x-forwarded-proto'] === 'https',
+      })
+      .json({ status: 'success', token });
+  } catch (err) {
+    res.json({ status: 'fail', error: err });
+  }
```
JuliaShavyrova, line 36: can you please explain flow of getting access_token (how often it is requested, where exactly it is stored, etc)?

ViliusGaizutis: It is called once, when the app loads in useEffect hook with empty [] dependencies array in the App component. And it is stored in redux store calling dispatch(setCredentials(res.data.token)); Although, now I think I probably could call hook on [auth.token] (i.e. useAppSelector(state => state.auth)) dependency?

JuliaShavyrova: ok, 2 more questions:
1. why do you use cookie-parser library?
2. if your app will be opened in two separate tabs, will it generate tokens twice?

ViliusGaizutis: 2. I guess it will. Since I'm not storing token in localStorage, as it is not a good idea, two separate tabs would work as two separate apps and on refresh would generate token, and it would not matter if I check whether token exists in redux store. And I can't read cookie, as it is set "httpOnly: true".
1. I am not actually, but maybe I should use it. Maybe I should not send token as a response in handleLogin fc on frontend, but call another fc, say checkAuthState, where I would do another request to server, and in the server I would use cookieParser to get req.cookies and then send the token?

client/src/features/auth/authSlice.ts
```diff
@@ -1,26 +1,30 @@
...
 export const authSlice = createSlice({
   name: 'auth',
   initialState,
-  reducers: {
-    setCredentials: (state, action: PayloadAction<string>) => {
-      state.token = action.payload;
-    }
+  reducers: {},
+  extraReducers: builder => {
+    builder.addMatcher(authApi.endpoints.authenticate.matchFulfilled, (state, action) => {
```
JuliaShavyrova, line 22: please add type definitions for state and action

JuliaShavyrova (approved):

PR #281, opened 2022-09-01
Add more info about weather and refactor

AndrewHelix:


src/types/weatherTypes.ts
```diff
@@ -0,0 +1,79 @@
...
+  sunsetEpoch: number;
+  moonrise: string;
+  moonset: string;
+  moonPhase: number;
+  uvIndex: number;
+  minVisibility: number;
+  pressure: number;
+  confidence: string;
+  solarRadiationSum: number;
+}[];
```
JuliaShavyrova: looks like these interfaces have duplicative properties. please move them into separate interface

src/components/main/Main.tsx
```diff
@@ -4,37 +4,33 @@ import commonStyle from '../../styles/commonStyles.css';
...
+  const [dailyWeather, setDailyWeather] = useState<DailyWeatherType>([]);
 
   async function fetchData(lon: number, lat: number) {
-    await getCity(lon, lat).then(locationData => setCity(locationData.city));
-    await getCityWeather(lon, lat).then(cityWeather => setWeather(cityWeather));
+    Promise.all([
+      getCity(lon, lat),
+      getWeather<CurrentWeatherType>('/api/v1/current/', lon, lat),
+      getWeather<{ forecast: HourlyWeatherType }>('/api/v1/forecast/hourly/', lon, lat),
+      getWeather<{ forecast: DailyWeatherType }>('/api/v1/forecast/daily/', lon, lat)
```
JuliaShavyrova: why do you store in state HourlyWetherType and DailyWeatherType as plain objects, but CurrentWeatherType is stored as object with next structure? do you need `current` property?
```
 {
  current: {
    ...
  }
}
```

src/components/main/MainCard.tsx
```diff
@@ -0,0 +1,55 @@
...
+    date: date.getDate(),
+    month: date.getMonth() + 1,
+    year: date.getFullYear(),
+    hours: date.getHours(),
+    minutes: date.getMinutes()
+  };
+};
+
+const MainCard: FC<MainCardProps> = ({ currentWeather, location }) => {
+  const date = convertTime(currentWeather.current.time);
```
JuliaShavyrova: the same question, why not just pass `currentWeather.current` as prop?

src/components/main/defaultStates.ts
```diff
@@ -0,0 +1,35 @@
+export const defaultCurrentWeather = {
```
JuliaShavyrova: need type definition here

src/components/main/Main.tsx
AndrewHelix: No, i don't need `current`
My bad. I'll fix it )

src/components/main/defaultStates.ts
AndrewHelix: ok:)
but if I have default state, which don't mathes with type, which defined in generic (like this `useState<CurrentWeatherType>(defaultCurrentWeather)`) I will have a mistake and can't compilate project.

src/components/main/MainCard.tsx
```diff
@@ -21,7 +21,7 @@ const convertTime = (timeData: string) => {
 };
 
 const MainCard: FC<MainCardProps> = ({ currentWeather, location }) => {
```
JuliaShavyrova, line 23: not required to be fixed in current PR, but would be better to use destructuring for `currentWeather` prop in this case

JuliaShavyrova (approved):

src/components/main/defaultStates.ts
JuliaShavyrova: that's why we are using typescript - to catch such errors :)

PR #286, opened 2022-09-07
Add context and reducer

AndrewHelix:
add context for city position info and change city, when I click on other city after search. In the future will be use context and for theme
In context use reduce for get state for value prop and change this state in other components


src/context/context.tsx
```diff
@@ -0,0 +1,42 @@
...
+export const dataContext = createContext(initState);
+
+export const ContextProvider = ({ children }: { children: ReactNode }) => {
+  const { position, error } = usePosition();
+
+  const { Provider } = dataContext;
+
+  const [state, dispatch] = useReducer(reducer, initState);
+
+  state.changePosition = (lat, los) => {
```
JuliaShavyrova: state should not be mutable. you don't need "changePosition" function inside state at all, make it a separate function and pass with context value

src/context/context.tsx
```diff
@@ -0,0 +1,42 @@
+import React, { createContext, ReactNode, useEffect, useReducer } from 'react';
+import { usePosition } from 'hooks/usePosition';
+import { reducer } from './reducer';
+import { CHANGE_POSITION } from './actions';
+
+export const initState = {
```
JuliaShavyrova: please add type definition

src/context/context.tsx
```diff
@@ -0,0 +1,42 @@
...
+};
+
+export const dataContext = createContext(initState);
+
+export const ContextProvider = ({ children }: { children: ReactNode }) => {
+  const { position, error } = usePosition();
+
+  const { Provider } = dataContext;
+
+  const [state, dispatch] = useReducer(reducer, initState);
```
JuliaShavyrova: also please add type definition

src/context/context.tsx
```diff
@@ -2,26 +2,49 @@ import React, { createContext, ReactNode, useEffect, useReducer } from 'react';
...
-  position: {
-    latitude: 0,
-    longitude: 0
+type InitStateType = {
+  state: {
+    position: {
+      latitude: number;
+      longitude: number;
+    };
+    positionError: string;
```
JuliaShavyrova: 1. you can reuse already created `State` interface here
2. i'd recommend not to use common names like State/Action/etc, make them more specific, like PositionContextValues/PositionContextState to avoid duplicating names across the app. also `InitStateType` is not an initial state according to its sense, it's just a type for context values, please rename to see the difference between `State` and `InitStateType` interfaces, because now it looks like both of them are used to describe the same data, but it's not true

src/context/context.tsx
```diff
@@ -26,23 +20,22 @@ export const initState: InitStateType = {
   changePosition: (lat: number, los: number) => {}
 };
 
-export const dataContext = createContext<InitStateType>(initState);
+export const dataContext = createContext<InitContextValuesType>(initValues);
```
JuliaShavyrova, line 23: `InitContextValuesType` still not an "init" type, it's just type for context :)

JuliaShavyrova (approved):

PR #287, opened 2022-09-08
Started working on dashboard

ViliusGaizutis:


client/src/common/components/ButtonSwitch/styles.tsx
```diff
@@ -0,0 +1,97 @@
...
+  svg:last-of-type {
+    right: 0.57rem;
+  }
+
+  span {
+    font-weight: 500;
+    transition: color 1.2s;
+  }
+
+  span:first-of-type {
```
JuliaShavyrova, line 72: why not create unified selector for color like this?
`span:first-of-type, span:last-of-type {  color: }`
`span:first-of-type { left: }`
`span:last-of-type { right: }`

client/src/common/components/Dashboard/DashboardCurrent/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const { query } = useAppSelector(state => state.location);
+
+  const { data, isLoading, isError } = useGetCurrLocationWeatherQuery(query, {
+    skip: !query
+  });
+
+  return (
+    <S.DashboardCurrent>
+      <h3>Current weather</h3>
+      <Widget color="primary" data={data && data.current} isLoading={isLoading} isError={isError} />
```
JuliaShavyrova: is <Widget/> a shared component that will be used to show data from different queries?

client/src/common/components/GeoLocation/index.tsx
```diff
@@ -14,24 +19,40 @@ import * as S from './styles';
...
-        {(posError || infoError) && !data && <p>No data available</p>}
-        {data && (
+        {data ? (
           <p>
             <span>{data.name}</span>, {data.country}
           </p>
+        ) : (
+          <>
+            {(posLoading || infoLoading) && (!posError || !infoError) && <p>Loading data...</p>}
+            {(posError || infoError) && (!posLoading || !infoLoading) && <p>No data available</p>}
```
JuliaShavyrova: if you have case when loading and error messages appear on the screen at the same time, use ternary operator to avoid it (you can extend your current one with `data` condition)

ViliusGaizutis: You mean nested ternary operator or what? Because if I do it that way, I get eslint error https://eslint.org/docs/latest/rules/no-nested-ternary. That's why before I did create a separate variable.

client/src/common/components/Dashboard/DashboardCurrent/index.tsx
ViliusGaizutis: No, this one just shows current weather. DashboardForecast (or maybe I'll give it a different name) component will be used to show data from different queries, like hourly, three-hourly, daily and air quality.

client/src/common/components/ButtonSwitch/styles.tsx
ViliusGaizutis: Ok, will do.

client/src/common/components/Dashboard/DashboardCurrent/index.tsx
JuliaShavyrova: in this case why do you pass weather data to `Widget` through props? you could just use `useGetCurrLocationWeatherQuery` inside it

client/src/common/components/GeoLocation/index.tsx
JuliaShavyrova: doesn't it work? (conditions order can be flexible)
``` 
{data ? (
 ...
) : loading ? (
  ...
) : error ? (
  ...
) : null
```

client/src/common/components/Dashboard/DashboardCurrent/index.tsx
ViliusGaizutis: Widget is a shared component, so depending on different data from props (currWeather, hourlyWeather, etc.) would render different results. For instance, in DashboardForecast I may have a list of widgets that each would show a specific hour's data from array of hourly data, which I would get from, say, useGetLocationHourlyWeatherQuery.

client/src/common/components/GeoLocation/index.tsx
ViliusGaizutis: I get "Do not nest ternary expressions.eslint[no-nested-ternary](https://eslint.org/docs/rules/no-nested-ternary)", if I do that.

JuliaShavyrova: you can move all this code to separate function and use if/else to return correct jsx fragment. it still will be easier to read than chain like this: 
`{(posError || infoError) && (!posLoading || !infoLoading) && <p>No data available</p>}`

client/src/common/components/GeoLocation/utils.tsx
```diff
@@ -0,0 +1,29 @@
+/* global GeolocationPositionError */
+import React, { ReactNode } from 'react';
+
+// Types
+import { LocationInfo } from '../../../types';
+
+export const renderConditionalJSX = (
```
JuliaShavyrova: function definitely should be renamed :) it's not just any conditional jsx, your function, first of all, returns location info (otherwise request status) and name of your function should show it

client/src/common/components/GeoLocation/utils.tsx
```diff
@@ -0,0 +1,29 @@
+/* global GeolocationPositionError */
```
JuliaShavyrova: comment needed?

client/src/common/components/GeoLocation/index.tsx
```diff
@@ -44,16 +46,7 @@ const GeoLocation: FunctionComponent = () => {
...
-          <p>
-            <span>{data.name}</span>, {data.country}
-          </p>
-        ) : (
-          <>
-            {(posLoading || infoLoading) && (!posError || !infoError) && <p>Loading data...</p>}
-            {(posError || infoError) && (!posLoading || !infoLoading) && <p>No data available</p>}
-          </>
-        )}
+        {renderConditionalJSX(data, posLoading || infoLoading, posError || infoError)}
```
JuliaShavyrova: 👍

client/src/common/components/GeoLocation/utils.tsx
ViliusGaizutis: Yes, otherwise I get an error. So I looked into it, and this was one of the solutions.

JuliaShavyrova: what is the error?

ViliusGaizutis: Eslint throws "'GeolocationPositionError' is not defined." I've searched quite a lot about it, but could not understand what I have to do. I've read like it should be available by default or something.

1. https://stackoverflow.com/questions/70759484/geolocationpositionerror-is-not-defined
2. https://stackoverflow.com/questions/69345470/how-can-i-bypass-eslint-calling-the-geolocation-interface-undefined-in-typescrip
3. https://stackoverflow.com/questions/67870306/geolocationpositionerror-code-1-message-user-denied-geolocation-prompt
There we're more stuff, but I don't remember where I've read about it.

client/src/common/components/GeoLocation/utils.tsx
ViliusGaizutis: Would renderDataOrReqStatus be better?

client/src/common/components/GeoLocation/utils.tsx
JuliaShavyrova: this error breaks your project compilation command or it is shown as warning in vscode?

ViliusGaizutis: Just checked it. It compiles successfully. Just shows up as red error warning in vscode, and marks files that depend on it as red. I don't like it :D

client/src/common/components/GeoLocation/utils.tsx
JuliaShavyrova: `Data` means it  can be any data, but in your case it should be location data. `renderLocationName\renderLocationLabel` is enough

client/src/common/components/GeoLocation/utils.tsx
JuliaShavyrova: you can find reason here, so no need to fix it in code, if you want - fix in vscode:
https://stackoverflow.com/a/66005573

client/src/common/components/GeoLocation/utils.tsx
ViliusGaizutis: But I plan to reuse it. So `Data` could be `LocationName` or `LocationCurrWeather`, or `LocationHourlyWeather`, or maybe even `LocationMap` (if I would manage to insert any) or something. So maybe `renderLocationData` then?

client/src/common/components/GeoLocation/utils.tsx
ViliusGaizutis: Yes, I remember seeing this comment. I have typescript@4.6.4. Can I leave that comment for the time being, so that eslint would shush?

JuliaShavyrova: you can leave it with comment that it's an IDE error and promise to configure your vscode :)

client/src/common/components/GeoLocation/index.tsx
```diff
@@ -46,7 +46,17 @@ const GeoLocation: FunctionComponent = () => {
     <S.GeoLocation>
       <Flex>
         <IconLocation />
-        {renderConditionalJSX(data, posLoading || infoLoading, posError || infoError)}
+        <RequestDataWrapper
+          data={data}
+          loading={posLoading || infoLoading}
+          error={posError || infoError}
+        >
+          {data && (
```
JuliaShavyrova: no need to use condition `data && ...` here, you already have this check inside `RequestDataWrapper`

ViliusGaizutis: I get and error `Uncaught TypeError: Cannot read properties of undefined (reading 'name')`, if I omit `data &&`.

JuliaShavyrova: you can use this syntax - `<span>{data?.name}</span>`, please read about it

ViliusGaizutis: Oh, I've actually used this syntax in this component already (`${pos?.lon}, ${pos?.lat}`). I guess I did not fully understand how it works. Will read more. Thank you

JuliaShavyrova (approved):

PR #289, opened 2022-09-09
Add night theme and info in countries page

AndrewHelix:
Screenshot 2022-09-09 175524
Screenshot 2022-09-09 175548
Screenshot 2022-09-09 175616


src/components/countries/CitySmallCard.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const [currentWeather, setCurrentWeather] = useState<CurrentWeatherType>(defaultCurrentWeather);
+
+  useEffect(() => {
+    getWeather<{ current: CurrentWeatherType }>('/current/', lon, lat).then(res =>
+      setCurrentWeather(res.current)
+    );
+  }, []);
+  return (
+    <>
+      {currentWeather.time ? (
```
JuliaShavyrova: why do you need to check `time` property here? can't you set `currentWeater` undefined as default and use condition `currentWeather ? (...)` ?

src/components/countries/Countries.tsx
```diff
@@ -1,11 +1,27 @@
...
+import { LocationInfoType } from 'types/cityInfoType';
+import { CitySmallCards } from './CitySmallCards';
 
 import commonStyle from '../../styles/commonStyles.css';
+import styles from './Countries.css';
 
 export const Countries: FC = () => {
+  const sortedByCountries: LocationInfoType[] = getFavoriteCities().sort((a, b) =>
+    a.country > b.country ? 1 : -1
+  );
```
JuliaShavyrova: please move sorter function to some `utils` file

src/components/main/MainCard.tsx
```diff
@@ -26,6 +28,32 @@ const MainCard: FC<MainCardProps> = ({ currentWeather, location }) => {
...
+  const [isFavorite, setIsFavorite] = useState(getIsFavorite);
+
+  const toggleFavoriteCities = () => {
+    if (!isFavorite) {
+      const favoriteCities = getFavoriteCities().concat(location);
+      setInLocalStorage(favoriteCities, 'favorite-cities');
+    } else {
+      const favoriteCities = getFavoriteCities().filter(city => city.id !== location.id);
+      setInLocalStorage(favoriteCities, 'favorite-cities');
+    }
```
JuliaShavyrova, line 48: please refactor function:
1. no matter if city is favorite or not, you will use `getFavoriteCities` function to get it, so it can be called once
2. no matter if city is favorite or not, you will use `setInLocalStorage` to set new list, so it also can be called once

src/components/main/MainCard.tsx
```diff
@@ -26,6 +28,32 @@ const MainCard: FC<MainCardProps> = ({ currentWeather, location }) => {
 
   const date = convertTime(time);
 
+  const getIsFavorite = (): boolean => {
+    const favoriteCities: LocationInfoType[] = getFavoriteCities();
+    return favoriteCities.some(city => city.id === location.id);
+  };
```
JuliaShavyrova: function can be moved out of component body

src/services/localStorage.ts
```diff
@@ -12,3 +14,9 @@ export function getFromLocalStorage(key: string): string {
   }
   return data;
 }
+
+export const getFavoriteCities = (): LocationInfoType[] => {
+  const favoriteCities: LocationInfoType[] = (getFromLocalStorage('favorite-cities') ||
```
JuliaShavyrova: 'favorite-cities' string can be defined as separate exported const

src/components/main/MainCard.tsx
AndrewHelix: In function I use location object. If I move this function, then I need to send location in parameter. But I use this function, when defined "isFavorite" state. If I will call this function, when defined state (`getIsFavorite(location)`), then this function will call for every re-render. not only when componentDidMount.
I can move this function and update state isFavorite after mounting, but I think it excess render, which I can avoid

src/services/localStorage.ts
AndrewHelix: I think function the best variant, because I need to get actual info about favorite cities. I do it, when call function. But if I use const, data about favorite cities will be update only when app reload.

JuliaShavyrova: can you please share code snippet of what you've done? using variable instead of hardcoded string should not work like you're describing

AndrewHelix: In localStorage.ts
`export const favoriteCitiesConst: LocationInfoType[] = (getFromLocalStorage('favorite-cities') ||
    []) as LocationInfoType[]`

In MainCard.tsx
`const toggleFavoriteCities = () => {
    let favoriteCities = favoriteCitiesConst //there
    console.log(favoriteCities)
    if (!isFavorite) {
      favoriteCities = favoriteCities.concat(location);
    } else {
      favoriteCities = favoriteCities.filter(city => city.id !== location.id);
    }
    setInLocalStorage(favoriteCities, 'favorite-cities');
  };`

for all click in btn I have same favoriteCities.

JuliaShavyrova: at initial comment i meant to store separately label for localStorage, not value you return from local storage:

```
export const FAVORITE_CITIES_LS_LABEL = 'favorite-cities';
export const getFavoriteCities = (): LocationInfoType[] => {
  const favoriteCities: LocationInfoType[] = (getFromLocalStorage(FAVORITE_CITIES_LS_LABEL) ||
```

AndrewHelix: I don't know why I didn't understand you right away 😅
Sorry:)

src/components/main/Main.tsx
```diff
@@ -37,7 +37,7 @@ const Main: FC = () => {
 
   useEffect(() => {
     if (position.longitude) {
-      fetchData(position.longitude, position.latitude);
+      //fetchData(position.longitude, position.latitude);
```
JuliaShavyrova, line 40: please remove if not needed

JuliaShavyrova (approved):

PR #290, opened 2022-09-11
task1/created main page

Silvik95:


src/UI/button/MyButton.jsx
```diff
@@ -0,0 +1,9 @@
+import styles from './MyButton.scss';
+
+export function MyBytton({ children, ...props }) {
```
virzhinia-petkova-redshelf: I feel like simply 'Button' would probably be a better fit.

virzhinia-petkova-redshelf: Also, lets stick to default exports at the end of the file like in the App component. If you were to add default props or prop types to your component, you'd have to refactor. You can read about it here https://reactjs.org/docs/typechecking-with-proptypes.html#function-components

src/UI/button/MyButton.jsx
```diff
@@ -0,0 +1,9 @@
...
+  return (
+    <button {...props} className={styles.myBtn}>
+      {children}
```
virzhinia-petkova-redshelf: since this is just a button if there a reason you're passing children here? What do you think about simply adding a prop to pass down some text/icon, etc

src/App.jsx
```diff
@@ -1,5 +1,15 @@
+import MainPage from './components/MainPage';
+import MyFooter from './header&footer/MyFooter';
```
virzhinia-petkova-redshelf: So, I don't think having a folder for the header and footer specifically is optimal. What do you think about simply adding the components to the /components folder?
here are some popular ways to group components: **https://reactjs.org/docs/faq-structure.html**

Make sure you aren't putting too much things at the root level of your app as it will get difficult to navigate as it grows.

src/UI/button/MyButton.scss
```diff
@@ -0,0 +1,10 @@
+.myBtn {
+  padding: 5px 15px;
+  color: rgb(0, 173, 173);
```
virzhinia-petkova-redshelf: what do you think about adding variables for colors?

src/UI/button/MyButton.jsx
```diff
@@ -0,0 +1,9 @@
+import styles from './MyButton.scss';
```
virzhinia-petkova-redshelf: Do you think this folder should also go into /components?

src/components/MainPage.jsx
```diff
@@ -0,0 +1,130 @@
+import React, { useState } from 'react';
+import styles from '../styles.scss';
+import { MyBytton } from '../UI/button/MyButton';
+import MyInput from '../UI/input/MyInput';
+
+export default function MainPage() {
+  const [searchQuery, setSearchQuery] = useState('');
+  const [items, setItems] = useState([
+    'Afghanistan',
```
virzhinia-petkova-redshelf: do you think extracting this array into a variable and placing is in a separate file would benefit your component? It will drastically decrease the size and you'd be able to concentrate on the actual logic of it.

src/components/MainPage.jsx
```diff
@@ -0,0 +1,130 @@
...
+    'Iceland',
+    'India'
+  ]);
+  const filteredItems = items.filter(item =>
+    item.toLocaleLowerCase().includes(searchQuery.toLocaleLowerCase())
+  );
+  return (
+    <div className={styles.MainPage}>
+      <div className={styles.search}>
+        <p>Find the right location</p>
```
virzhinia-petkova-redshelf, line 17: should this go into a constant?

src/components/MainPage.jsx
```diff
@@ -0,0 +1,130 @@
...
+        <MyInput
+          value={searchQuery}
+          onChange={e => setSearchQuery(e.target.value)}
```
virzhinia-petkova-redshelf, line 20: it looks like you're always going to be using those two in the input component. Is it worth adding the props explicitly to it to provide clarity on how the component is supposed to be used?

src/header&footer/MyFooter.jsx
```diff
@@ -0,0 +1,20 @@
+import { useEffect, useState } from 'react';
+import styles from './header&footer.scss';
+
+export default function MyFooter() {
+  const [currentTime, setCurrentTime] = useState(new Date().toLocaleTimeString());
+  const interval = setInterval(() => {
+    setCurrentTime(new Date().toLocaleTimeString());
+  }, 1000);
```
virzhinia-petkova-redshelf: setInterval is a side effect and all side effects should be wrapped in a useEffect. Please move it there. 

Further reading: https://reactjs.org/docs/hooks-effect.html

src/header&footer/MyFooter.jsx
```diff
@@ -0,0 +1,20 @@
...
+  useEffect(() => {
+    return () => {
+      console.log('test');
```
virzhinia-petkova-redshelf: did you forget this :)

src/header&footer/MyFooter.jsx
```diff
@@ -0,0 +1,20 @@
...
+      clearInterval(interval);
+    };
+  }, []);
+  return (
+    <div className={styles.appfooter}>
+      <h3>{currentTime}</h3>
```
virzhinia-petkova-redshelf: I know that accessibility is not part of the training but it's important to understand that html elements should be used like they're supposed to so you're not confusing people using assistive technology like screen readers.

If your component is not actually acting like a heading, then just use a `<p>` tag and style it the way you want it.

src/header&footer/MyFooter.jsx
```diff
@@ -0,0 +1,20 @@
+import { useEffect, useState } from 'react';
+import styles from './header&footer.scss';
```
virzhinia-petkova-redshelf: you can have separate style files for header & footer since they are separate components

src/header&footer/header&footer.scss
```diff
@@ -0,0 +1,28 @@
...
+  display: flex;
+  align-items: center;
+  position: fixed;
+  bottom: 0;
+  width: 100%;
+  h3 {
+    margin: 0;
+    padding-left: 90px;
+    color: rgb(221, 170, 17);
+  }
```
virzhinia-petkova-redshelf: since you're repeating those styles why don't you think of a way to only have them in one place

virzhinia-petkova-redshelf (commented):
Overall good job. :)

I've left some comments.

Also, please make sure you provide a brief description of what you've done in the PR when opening it.

src/components/Footer.jsx
```diff
@@ -0,0 +1,20 @@
...
+  useEffect(() => {
+    const interval = setInterval(() => {
+      setCurrentTime(new Date().toLocaleTimeString());
+    }, 1000);
+    return () => {
+      clearInterval(interval);
+    };
+  }, []);
+  return (
+    <div className={styles.appfooter}>
```
virzhinia-petkova-redshelf, line 15: is there a reason some of your styles are capitalized and some arent?

src/components/Header.jsx
```diff
@@ -0,0 +1,9 @@
+import styles from '../styles.scss';
+function Header() {
+  return (
+    <div className={styles.appheader}>
+      <h3>CanYouSwim??</h3>
```
virzhinia-petkova-redshelf: can you update the h3 and also move this to a constant

src/components/MainPage.jsx
virzhinia-petkova-redshelf: thoughts here?

src/components/MainPage.jsx
virzhinia-petkova-redshelf: thoughts here?

src/components/MainPage.jsx
virzhinia-petkova-redshelf: thoughts here?

src/UI/button/Button.jsx
```diff
@@ -0,0 +1,10 @@
+import styles from './Button.scss';
+
+function Button(props) {
```
virzhinia-petkova-redshelf, line 3: you can destructure props like
```suggestion
function Button({ text, ...props}) {
```

that way it's clear what the component receives and prevents some console errors

src/UI/button/Button.jsx
```diff
@@ -0,0 +1,10 @@
...
+  return (
+    <button {...props} className={styles.Btn}>
```
virzhinia-petkova-redshelf: spreading the props should usually be done last unless you specifically want something not to get overwritten

src/UI/button/Button.scss
```diff
@@ -0,0 +1,10 @@
+.Btn {
+  padding: 5px 15px;
+  color: rgb(0, 173, 173);
```
virzhinia-petkova-redshelf: constant? you can create a palette and use it everywhere. If you have to switch themes it will be very difficult for you as it will require a ton of refactoring

src/styles.scss
```diff
@@ -0,0 +1,63 @@
+@import url('https://fonts.googleapis.com/css2?family=Ubuntu:wght@300;400;700&display=swap');
+body {
+  font-family: 'Ubuntu', sans-serif;
+  margin: 0 auto;
+}
+$bgc: teal;
+$headerAndFotterColor: rgb(221, 170, 17);
+$searchItemsColor: rgb(23, 203, 216);
```
virzhinia-petkova-redshelf: say you want to use teal for a button border. Won't it be confusing for the variable to say bgc? YOu can take a look at this article here https://medium.com/@stefanmorris/proper-naming-of-color-variables-683264fec0f4

src/components/MainPage.jsx
Silvik95: what I need to put into variable?)

src/components/MainPage.jsx
Silvik95: can you give me little hint, how you see this

src/components/MainPage.jsx
virzhinia-petkova-redshelf: it's a long topic but here are some good reasons: https://softwareengineering.stackexchange.com/questions/365339/what-is-wrong-with-magic-strings

I understand that we're working on a small app but if you were to expand it in the future, it'll be much easier for you to do things like translation for example if you have all your magic strings into constants. It's a good habit to have.

src/components/MainPage.jsx
virzhinia-petkova-redshelf: you can pass value & onChange as props in the input component itself like 

`const Input = ({ value, onChange, ...props })`...

it provides clarity on what the component needs to function properly

src/components/MainPage.jsx
virzhinia-petkova-redshelf: so you decided against moving this to a constant?

Silvik95: sorry, I didn't see your comment. I read this topic and did not understand for what need past static code into constant. How it can be reusable?

virzhinia-petkova-redshelf: The point is that your life will probably be easier in the future. Another example is that you could potentially use the string in UI tests. You'd need to reuse it then.

Anyway, it's not something critical right now so you can leave it.

virzhinia-petkova-redshelf (approved):

PR #291, opened 2022-09-12
Reverted back to Redux + Thunk and changed folder structure

ViliusGaizutis:


client/src/redux/reducers/theme.ts
```diff
@@ -0,0 +1,24 @@
+import { ActionType, Action } from '../actionTypes/theme';
+
+interface ThemeState {
+  theme: string;
+}
+
+const initialState = {
+  theme: 'light'
```
AndrewHelix: I think in this place you can write `initialState = 'light'`
Then you can use in app `state.theme`, but not `state.theme.theme` like in Header and Input (maybe in other components)

client/src/redux/reducers/location.ts
```diff
@@ -0,0 +1,155 @@
...
+    loading: false,
+    error: null
+  },
+  info: {
+    data: null,
+    loading: false,
+    error: null
+  },
+  weather: {
+    current: {
```
AndrewHelix, line 37: Do you need use 'current'. I think you can save weather data without 'current'
`weather: {
  data: ...
}`

client/src/components/Widget/Widget.tsx
```diff
@@ -23,40 +23,40 @@ const Widget: FunctionComponent<WidgetProps> = ({ color, data }) => {
...
+                <p>{capitalize(data?.symbolPhrase)}</p>
               </Flex>
             </S.WidgetImg>
             <S.WidgetDetailsItem color={color}>
-              <span>Cloudiness</span> {data.cloudiness}%
+              <span>Cloudiness</span> {data?.cloudiness}%
             </S.WidgetDetailsItem>
             <S.WidgetDetailsItem color={color}>
-              <span>Precipation</span> {data.precipProb}%, {data.precipRate}&nbsp;mm/h
+              <span>Precipation</span> {data?.precipProb}%, {data?.precipRate}&nbsp;mm/h
```
AndrewHelix, line 59: Why you use so many `data?.`
Maybe better will be 
`{data ? <S.Widget color={color}>...</ S.Widget> : <p>some error or something else</p> }`

client/src/redux/actionCreators/location.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+  }
+};
+
+export const getLocationCurrWeather = (query: string) => async (dispatch: Dispatch<Action>) => {
+  dispatch({ type: ActionType.GET_LOCATION_CURRENT_WEATHER_PENDING });
+
+  try {
+    const data = await axios.get(
+      `http://localhost:3000/forecaApi/current/${query}&tempunit=C&windunit=MS`,
```
AndrewHelix: Why you don't use base url in const? 
Example: const BASE_USE = 'http://localhost:3000/forecaApi/'
I think it will be better

client/src/redux/reducers/location.ts
ViliusGaizutis: There will be other properties for `hourly`, `three-hourly`, etc., so that's why this one is called `current`.

client/src/redux/actionCreators/location.ts
ViliusGaizutis: I will, but later. That whole function needs refactoring, because all these actionCreators have similar and/or duplicate code.

client/src/components/Widget/Widget.tsx
ViliusGaizutis: Actually, in the beginning I've had something similar, i.e. `{data && <S.Widget color={color}>...</ S.Widget>}`, but changed to `data?.x` syntax. I don't know if there is much difference here, do you?

client/src/redux/reducers/theme.ts
ViliusGaizutis: How do you propose to update it then in themeReducer function?

client/src/components/Widget/Widget.tsx
AndrewHelix: I think `{data && <S.Widget color={color}>...</ S.Widget>}` looks more beautiful

client/src/redux/reducers/theme.ts
AndrewHelix: Maybe return `return action.payload`.
I don't know. Maybe you are right

client/src/App.tsx
```diff
@@ -1,41 +1,41 @@
...
       user: process.env.USER,
       password: process.env.PASSWORD
     };
 
     if (!token) {
-      await authorize(credentials);
+      await dispatch(authorize(credentials));
     }
-  }, [authenticate, authorize]);
+  }, [dispatch]);
```
AndrewHelix, line 38: Why you use useCallback and 'dispatch' in dependencies array here? 
You are not passing a callback using dispatch to a child component. 
'react-redux' documentation says that "The dispatch function reference will be stable as long as the same store instance is being passed to the <Provider>. Normally, that store instance never changes in an application."
I think 'handleAuth' will not change for all life app because App component will not update. 
I think you don't should to use useCallback and call handleAuth in useEffect witch empty dependencies array

client/src/components/Search/Search.tsx
```diff
@@ -1,46 +1,59 @@
...
+  const dispatch = useDispatch();
+
   // Handlers
   const handleInputChange = (e: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>): void => {
     setSearchQuery(e.currentTarget.value);
   };
   const handleInputFocus = (): void => setListShown(true);
   const handleClickOutside = (): void => setListShown(false);
+  const handleSearchLocations = useCallback(() => {
+    if (debouncedSearchQuery !== '') dispatch(searchLocations(debouncedSearchQuery));
```
JuliaShavyrova: please make condition a bit shorter, no need to compare exactly with empty string :)

client/src/components/Widget/Widget.tsx
JuliaShavyrova: both ways are valid :)

client/src/redux/actionCreators/location.ts
JuliaShavyrova: agree with it, please move url to separate variable

client/src/redux/actionTypes/auth.ts
```diff
@@ -0,0 +1,27 @@
...
+  type: ActionType.AUTHORIZE_SUCCESS | ActionType.AUTHENTICATE_SUCCESS;
+  payload?: AccessToken;
+}
+
+interface ActionFail {
+  type: ActionType.AUTHORIZE_FAIL | ActionType.AUTHENTICATE_FAIL;
+  payload: string;
+}
+
+export type Action = ActionPending | ActionSuccess | ActionFail;
```
JuliaShavyrova, line 27: 👍

client/src/redux/reducers/auth.ts
```diff
@@ -0,0 +1,62 @@
...
+      };
+    }
+    case ActionType.AUTHENTICATE_SUCCESS: {
+      return {
+        ...state,
+        loading: false,
+        accessToken: action.payload
+      };
+    }
+    case ActionType.AUTHENTICATE_FAIL: {
```
JuliaShavyrova: looks like cases ActionType.AUTHORIZE_FAIL and ActionType.AUTHENTICATE_FAIL have the same state returned, please optimize `switch/case` statement

client/src/redux/reducers/theme.ts
JuliaShavyrova: agree with Andrew comment, `state.theme` can be a simple string

client/src/types/index.ts
```diff
@@ -4,11 +4,15 @@ export interface AuthorizationRequest {
...
-  status: 'success' | 'fail';
+  data: {
+    status: 'success' | 'fail';
+  };
 }
 export interface AuthenticationResponse {
-  status: 'success' | 'fail';
-  token: string;
+  data: {
+    status: 'success' | 'fail';
```
JuliaShavyrova:
```
{
  status: 'success' | 'fail';
};
```
it can be a separate interface

client/src/components/Widget/Widget.tsx
```diff
@@ -23,40 +23,40 @@ const Widget: FunctionComponent<WidgetProps> = ({ color, data }) => {
...
-            <S.WidgetTemp>{data.temperature}°</S.WidgetTemp>
+            <S.WidgetTemp>{data?.temperature}°</S.WidgetTemp>
             <S.WidgetDetailsItem color={color}>
-              <span>Real Feel</span> {data.feelsLikeTemp}°
+              <span>Real Feel</span> {data?.feelsLikeTemp}°
             </S.WidgetDetailsItem>
             <S.WidgetDetailsItem color={color}>
-              <span>Wind</span> {data.windDirString}, speed {data.windSpeed}&nbsp;m/s, gust{' '}
-              {data.windGust} m/s
+              <span>Wind</span> {data?.windDirString}, speed {data?.windSpeed}&nbsp;m/s, gust{' '}
```
AndrewHelix, line 38: Do you need `{' '}` ?

client/src/App.tsx
ViliusGaizutis: ESLint wants it to be in such a way, I get such warnings otherwise:
`React Hook useEffect has a missing dependency: 'handleAuth'. Either include it or remove the dependency array.eslint[react-hooks/exhaustive-deps]`(https://github.com/facebook/react/issues/14920)

`React Hook useCallback has a missing dependency: 'dispatch'. Either include it or remove the dependency array.eslint[react-hooks/exhaustive-deps]` (https://github.com/facebook/react/issues/14920)

And, actually, at the beginning when I used RTK query, I had it with an `[]` array, but since ESLint wanted it - I included `handleAuth` in the array (`[handleAuth]`), and then I had to call `useCallback` hook as well.

client/src/components/Widget/Widget.tsx
ViliusGaizutis: Apparently, yes. I need a space between `gust` and `{data.windGust} m/s`, but the latter goes to next line after file save, so Prettier adds this `{' '}`.

client/src/App.tsx
AndrewHelix: I think you can create handleAuth function without useCallback and call this function in useEffect with empty dependencies array. Then you get token, when app started. 
If I'm right, your App component doesn't update for all application life. That's why your handleAuth function will not create again and again. Consequently you don't need a hook useCallback.
But! I could be wrong. I think Julia can help find truth in our discussion :)

JuliaShavyrova: @ViliusGaizutis please find the answer and explain Andrew why `dispatch` needs to be mentioned in dependency array exactly in your case and in your project. Answer “eslint shows error otherwise” is not a correct developer answer, you need to understand why it works like this, you are a code owner :)

ViliusGaizutis: React Redux docs say: "The `dispatch` function reference will be stable as long as the same store instance is being passed to the `<Provider>`. Normally, that store instance never changes in an application.
However, the React hooks lint rules do not know that `dispatch` should be stable, and will warn that the `dispatch` variable should be added to dependency arrays for `useEffect` and `useCallback`. The simplest solution is to do just that."
(https://react-redux.js.org/api/hooks#usedispatch)

client/src/constants/index.ts
```diff
@@ -0,0 +1,3 @@
+export enum Constants {
+  BASE_URL = 'http://localhost:3000'
+}
```
JuliaShavyrova: no need to store base url as enum, simple variable in current file will be enough

JuliaShavyrova (approved):

PR #292, opened 2022-09-13
Add info in about page and redirect form countries page

AndrewHelix:
Screenshot 2022-09-13 172450


src/components/about/About.tsx
```diff
@@ -0,0 +1,42 @@
...
+          <p>
+            {' '}
+            On the favorites page, click on the desired city to view detailed weather in that city.{' '}
+          </p>
+          <p>
+            {' '}
+            On the main page with detailed weather information, select the number of days to see the
+            weather for the future.{' '}
+          </p>
+        </p>
```
ViliusGaizutis, line 30: As per https://www.w3.org/TR/html401/struct/text.html, "the `<p>` element represents a paragraph. It cannot contain [block-level elements] (including `<p>` itself), therefore you cannot nest `<p>` elements, it is illegal. Why not write one huge paragraph? If you need a new line, you can use `<br />` tag.

JuliaShavyrova (approved):

PR #293, opened 2022-09-15
Added functionality for getting hourly, three-hourly and daily weather

ViliusGaizutis:


ViliusGaizutis:
Screenshot 2022-09-15 143422
Home page layout design (light)

Screenshot 2022-09-15 143820
Expanded widget on click

Screenshot 2022-09-15 143648
Home page layout design (dark)

client/src/components/Dashboard/Dashboard.styles.tsx
```diff
@@ -24,8 +26,50 @@ export const DashboardCurrent = styled.div<StylesProps>`
...
+  color: ${props => {
+    if (!props.active && props.theme === 'light') {
+      return theme.palette.black;
+    }
+    if (!props.active && props.theme !== 'light') {
+      return theme.palette.white;
+    }
+    if (props.active) {
+      return theme.palette.primary.dark;
+    }
```
JuliaShavyrova, line 58: please refactor using if/else and nested conditions, will be easier to read :)

client/src/components/Dashboard/DashboardForecast/DashboardForecast.tsx
```diff
@@ -29,6 +143,19 @@ const DashboardForecast: FunctionComponent = () => {
...
         </Flex>
       </S.DashboardToolbar>
+      <S.DashboardForecastWidgets>
+        <RequestDataWrapper data={data} loading={loading} error={error}>
+          <Carousel setCarouselChildPointerEv={setCarouselChildPointerEv}>
+            {data?.map(item => (
+              <Widget
+                key={item.time || item.date}
+                data={item}
+                pointerEvents={carouselChildPointerEv ? 'true' : 'false'}
```
JuliaShavyrova: why 'true'/'false' are not boolean?

client/src/components/Dashboard/DashboardForecast/DashboardForecast.tsx
```diff
@@ -1,23 +1,137 @@
...
+  } = useAppSelector(state => state.location.weather.threeHourly);
+  const { data: dailyData, loading: dailyLoading, error: dailyError } = useAppSelector(
+    state => state.location.weather.daily
+  );
+
+  const [forecastType, setForecastType] = useState<ForecastTypeState>({
+    hourly: true,
+    threeHourly: false,
+    daily: false
+  });
```
JuliaShavyrova, line 50: as far as i understand your UI, you have 3 tabs with different forecast types and clicking on the type, it will show selected weather. according to this, some things can be refactored:
1. you don't need such a complex state, `selectedForecastType` will be enough
2. would be great to store forecast type labels as enum

client/src/components/Dashboard/DashboardForecast/DashboardForecast.tsx
```diff
@@ -1,23 +1,137 @@
...
+  if (forecastType.threeHourly) {
+    data = threeHourlyData;
+    loading = threeHourlyLoading;
+    error = threeHourlyError;
+  }
+  if (forecastType.daily) {
+    data = dailyData;
+    loading = dailyLoading;
+    error = dailyError;
+  }
```
JuliaShavyrova: too much duplicative code across the whole component. please refactor, choose what request to make, what selector to use, what label to show and other things according to `selectedForecastType` state value

client/src/components/Dashboard/DashboardForecast/DashboardForecast.tsx
```diff
@@ -1,23 +1,137 @@
...
 
   return (
     <S.DashboardForecast>
       <S.DashboardToolbar>
         <Flex justifySpaceBetween>
-          <div>Nowcast Hourly Three-hourly Daily</div>
+          <S.DashboardForecastTypes>
+            <Flex justifySpaceBetween>
+              <S.DashboardForecastType
```
JuliaShavyrova: these 3 components can be rendered using `map` method

client/src/components/RequestDataWrapper/RequestDataWrapper.tsx
```diff
@@ -4,9 +4,9 @@ import React, { FunctionComponent } from 'react';
 import { LocationInfo, WeatherInfo } from '../../types';
 
 interface RequestDataWrapperProps {
-  data: LocationInfo | WeatherInfo | null;
-  loading: boolean;
-  error: boolean | string | GeolocationPositionError | null;
+  data?: LocationInfo | WeatherInfo | WeatherInfo[] | null;
```
JuliaShavyrova: just use `unknown/any` if there is no ability to use generic

client/src/components/Widget/Widget.styles.tsx
```diff
@@ -1,56 +1,110 @@
...
+    if (props.color !== 'primary' && props.theme === 'light') {
+      return theme.palette.grey.light;
+    }
+    if (props.color !== 'primary' && props.theme !== 'light') {
+      return theme.palette.grey.darkest;
+    }
+    if (props.color === 'primary') {
+      return theme.palette.primary.medium;
+    }
+    return null;
```
JuliaShavyrova, line 56: the same about conditions

client/src/redux/actionCreators/location.ts
```diff
@@ -52,6 +52,65 @@ export const getLocationCurrWeather = (query: string) => async (dispatch: Dispat
...
+    dispatch({
+      type: ActionType.GET_LOCATION_DAILY_WEATHER_SUCCESS,
+      payload: data.data.forecast
+    });
+  } catch (err) {
+    if (err instanceof Error) {
+      dispatch({ type: ActionType.GET_LOCATION_DAILY_WEATHER_FAIL, payload: err.message });
+    }
+  }
+};
```
JuliaShavyrova: can these 3 functions be united to one or maybe one additional generic function could be created? i see that only url and action types are different

client/src/redux/actionTypes/location.ts
```diff
@@ -38,11 +53,29 @@ interface ActionCurrentWeatherSuccess {
...
+
+interface ActionThreeHourlyWeatherSuccess {
+  type: ActionType.GET_LOCATION_THREE_HOURLY_WEATHER_SUCCESS;
+  payload: WeatherInfo[];
+}
+
+interface ActionDailyWeatherSuccess {
+  type: ActionType.GET_LOCATION_DAILY_WEATHER_SUCCESS;
+  payload: WeatherInfo[];
+}
```
JuliaShavyrova, line 50: the same about one common interface here

client/src/components/Dashboard/DashboardForecast/DashboardForecast.tsx
ViliusGaizutis: At first, I tried passing `carouselChildPointerEv` as a boolean, but I got "Warning: Received `false` for a non-boolean attribute `pointerEvents`. If you want to write it to the DOM, pass a string instead: pointerEvents="false" or pointerEvents={value.toString()}. If you used to conditionally omit it with pointerEvents={condition && value}, pass pointerEvents={condition ? value : undefined} instead."

So I tried the last suggestion from the list above, i.e. `pointerEvents={carouselChildPointerEv || undefined}`, and still got this warning in console. The project is compiled successfully in terminal, though.

I also checked stackoverflow @ https://stackoverflow.com/questions/49784294/warning-received-false-for-a-non-boolean-attribute-how-do-i-pass-a-boolean-f and tried `{carouselChildPointerEv ? 1 : 0}` and `{carouselChildPointerEv ? true : undefined}`, but these did not work.

I should have looked at this comment @ https://stackoverflow.com/a/64157580, it's a styled-components issue :) Will fix it.

client/src/components/Dashboard/Dashboard.styles.tsx
ViliusGaizutis: I try to follow eslint rules. In this case, https://eslint.org/docs/latest/rules/no-nested-ternary:
if (foo) {
  thing = bar;
} else if (baz === qux) {
  thing = quxx;
} else {
  thing = foobar;
}
But if you remember from before, I also get warning "Unnecessary 'else' after 'return'.eslint[no-else-return](https://eslint.org/docs/rules/no-else-return)".

Also Igor Korshuk said not to use ternary expressions, especially nested, since it makes code implicit, difficult to read.

JuliaShavyrova: ok, looks like we have different opinions on that, you can follow eslint :)

client/src/components/Dashboard/DashboardForecast/DashboardForecast.tsx
```diff
@@ -1,23 +1,97 @@
...
+  const { data, loading, error } = useAppSelector(state => {
+    if (selectedForecastType === ForecastType.hourly) {
+      return state.location.weather.hourly;
+    }
+    if (selectedForecastType === ForecastType.threeHourly) {
+      return state.location.weather.threeHourly;
+    }
+    if (selectedForecastType === ForecastType.daily) {
+      return state.location.weather.daily;
+    }
```
JuliaShavyrova: you can use `switch` here

client/src/components/Dashboard/DashboardForecast/DashboardForecast.tsx
```diff
@@ -1,23 +1,97 @@
...
+
+    if (selectedForecastType === ForecastType.hourly) {
+      dispatch(getLocationHourlyWeather(query));
+    }
+    if (selectedForecastType === ForecastType.threeHourly) {
+      dispatch(getLocationThreeHourlyWeather(query));
+    }
+    if (selectedForecastType === ForecastType.daily) {
+      dispatch(getLocationDailyWeather(query));
+    }
```
JuliaShavyrova, line 74: also `switch` can be used

client/src/components/RequestDataWrapper/RequestDataWrapper.tsx
```diff
@@ -1,20 +1,17 @@
...
-  loading: boolean;
-  error: boolean | string | GeolocationPositionError | null;
+interface RequestDataWrapperProps<T> {
+  data?: T | T[];
+  loading?: boolean;
+  error?: boolean | string | GeolocationPositionError | null;
 }
 
-const RequestDataWrapper: FunctionComponent<RequestDataWrapperProps> = ({
+const RequestDataWrapper = <T,>({
```
JuliaShavyrova: wasn't you able to use `T extends unknown` or something like that?

client/src/components/Widget/Widget.styles.tsx
```diff
@@ -1,56 +1,109 @@
...
-  color: string;
+  color?: string;
+  active?: boolean;
+  pointerEvents?: string;
 }
 
 export const Widget = styled.article<WidgetStyles>`
-  width: 100%;
-  color: ${theme.palette.black};
+  width: ${(props: WidgetStyles) =>
```
JuliaShavyrova: `props` variable can be destructured here and in other places

client/src/components/Dashboard/DashboardForecast/DashboardForecast.tsx
```diff
@@ -29,6 +103,19 @@ const DashboardForecast: FunctionComponent = () => {
...
+            {data?.map(item => (
+              <Widget
+                key={item.time || item.date}
+                data={item}
+                pointerEvents={carouselChildPointerEv}
+              />
+            ))}
+          </Carousel>
+        </RequestDataWrapper>
+      </S.DashboardForecastWidgets>
```
JuliaShavyrova, line 123: component looks much better and shorter 👍

client/src/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,83 @@
...
+import * as S from './Carousel.styles';
+
+interface CarouselState {
+  isPressed: boolean;
+  startX: number | null;
+  x: number | null;
+}
+
+interface CarouselProps {
+  setCarouselChildPointerEv: (bool: boolean) => void;
```
JuliaShavyrova: would be better to rename `bool` variable so another developer could easily understand what exactly should be passed when calling the function

client/src/components/RequestDataWrapper/RequestDataWrapper.tsx
ViliusGaizutis: I don't know. I've read eslint rules again, and it says that it's unnecessary, because it is extended to unknown by default. (https://typescript-eslint.io/rules/no-unnecessary-type-constraint/). If I have to add here in this component, then would I have to add same syntax to other places where I've used generics as well?

JuliaShavyrova: you need to add something after <T to help compiler understand that it's a typescript syntax, not jsx. i still recommend to use `extends unknown` for this purpose because it doesn't cause guesses about haven't anything missed after comma sign

client/src/components/Dashboard/DashboardForecast/DashboardForecast.tsx
```diff
@@ -58,14 +60,21 @@ const DashboardForecast: FunctionComponent = () => {
...
+      case ForecastType.threeHourly: {
+        dispatch(getLocationThreeHourlyWeather(query));
+        break;
+      }
+      case ForecastType.daily: {
+        dispatch(getLocationDailyWeather(query));
+        break;
+      }
+      default:
+        dispatch(getLocationHourlyWeather(query));
```
JuliaShavyrova: first case and default case can be written like this:
```
case ForecastType.hourly: 
default:
        dispatch(getLocationHourlyWeather(query));
```

client/src/components/Dashboard/DashboardForecast/DashboardForecast.tsx
```diff
@@ -36,17 +36,19 @@ const DashboardForecast: FunctionComponent = () => {
...
+        return state.location.weather.hourly;
+      }
+      case ForecastType.threeHourly: {
+        return state.location.weather.threeHourly;
+      }
+      case ForecastType.daily: {
+        return state.location.weather.daily;
+      }
+      default:
+        return state.location.weather.hourly;
```
JuliaShavyrova: state can be destructured

JuliaShavyrova (approved):

PR #294, opened 2022-09-15
Task 1 (Creating main page - header, footer, cities list + filtering)

StefanAtanasov98:
Creating the main page of Weather App includes the following components/functionalities:

- Header with the name of the app;
- Footer with the current time;
- Navigation bar with a dropdown button containing different cities and an input field for filtering them;

app

### DIFF
--- a/src/components/Homepage/Homepage.js
+++ b/src/components/Homepage/Homepage.js
@@ -1,0 +1,37 @@
+import React, { useState, useEffect } from 'react';
+import * as homepageStyles from '../../styles/homepage.module.css';
+import { Nav } from './Nav/Nav.js';
+
+export function Homepage(props) {
+  const [time, setTime] = useState('Loading...');
+
+  useEffect(() => {
+    const interval = setInterval(() => setTime(getCurrentTime()), 1000);
+    return () => clearInterval(interval);
+  }, []);
+
+  const getCurrentTime = () => {
+    const date = new Date();
+    const hours = date.getHours() < 10 ? `0${date.getHours()}` : date.getHours();
+    const minutes = date.getMinutes() < 10 ? `0${date.getMinutes()}` : date.getMinutes();
+    const seconds = date.getSeconds() < 10 ? `0${date.getSeconds()}` : date.getSeconds();
+    const currentTime = `${hours}:${minutes}:${seconds}`;
+
+    return currentTime;
+  };
+
+  return (
+    <>
+      <header>
+        <h1>Weather Forecast</h1>
+      </header>
+      <section className="layout">
+        <Nav styles={homepageStyles} />
+        <main></main>
+      </section>
+      <footer>
+        <span>Current time: {time}</span>
+      </footer>
+    </>
+  );
+}
--- a/src/components/Homepage/Homepage.js
+++ b/src/components/Homepage/Homepage.js
@@ -1,37 +1,19 @@
 import React, { useState, useEffect } from 'react';
-import * as homepageStyles from '../../styles/homepage.module.css';
+import * as styles from '../../styles/HomePage.module.css';
+import { Header } from '../Header/Header.js';
+import { Footer } from '../Footer/Footer.js';
 import { Nav } from './Nav/Nav.js';
 
-export function Homepage(props) {
-  const [time, setTime] = useState('Loading...');
-
-  useEffect(() => {
-    const interval = setInterval(() => setTime(getCurrentTime()), 1000);
-    return () => clearInterval(interval);
-  }, []);
-
-  const getCurrentTime = () => {
-    const date = new Date();
-    const hours = date.getHours() < 10 ? `0${date.getHours()}` : date.getHours();
-    const minutes = date.getMinutes() < 10 ? `0${date.getMinutes()}` : date.getMinutes();
-    const seconds = date.getSeconds() < 10 ? `0${date.getSeconds()}` : date.getSeconds();
-    const currentTime = `${hours}:${minutes}:${seconds}`;
-
-    return currentTime;
-  };
-
+function HomePage(props) {
   return (
     <>
-      <header>
-        <h1>Weather Forecast</h1>
-      </header>
-      <section className="layout">
-        <Nav styles={homepageStyles} />
-        <main></main>
-      </section>
-      <footer>
-        <span>Current time: {time}</span>
-      </footer>
+      <Header />
+      <main>
+        <Nav />
+      </main>
+      <Footer />
     </>
   );
 }
+
+export { HomePage };
--- a/src/components/Homepage/Homepage.js
+++ b/src/components/Homepage/Homepage.js
@@ -1,17 +1,13 @@
 import React, { useState, useEffect } from 'react';
 import * as styles from '../../styles/HomePage.module.css';
-import { Header } from '../Header/Header.js';
-import { Footer } from '../Footer/Footer.js';
 import { Nav } from './Nav/Nav.js';
 
 function HomePage(props) {
   return (
     <>
-      <Header />
       <main>
         <Nav />
       </main>
-      <Footer />
     </>
   );
 }
--- a/src/components/Homepage/Nav/DropdownMenu.js
+++ b/src/components/Homepage/Nav/DropdownMenu.js
@@ -2,32 +2,54 @@ import React, { useState } from 'react';
 import * as styles from '../../../styles/DropdownMenu.module.css';
 
 const initialCitiesState = [
-  { name: 'Sofia', id: Math.random() },
-  { name: 'Athens', id: Math.random() },
-  { name: 'Istanbul', id: Math.random() }
+  { name: 'Sofia', id: 1 },
+  { name: 'Athens', id: 2 },
+  { name: 'Istanbul', id: 3 }
 ];
 
 function DropdownMenu(props) {
   const [isOpen, setIsOpen] = useState(false);
   const [cities, setCities] = useState(initialCitiesState);
+  const [searchedCity, setSearchedCity] = useState({});
 
   const showCities = ev => {
-    const dropdownContent = [...document.getElementsByClassName(styles.dropdownContent)][0];
     if (isOpen) {
       setIsOpen(false);
-      dropdownContent.style.display = 'none';
     } else {
       setIsOpen(true);
-      dropdownContent.style.display = 'inline-block';
     }
   };
 
-  const citiesList = cities.map(city => <a key={city.id} href="#">{city.name}</a>);
+  const onSearch = ev => {
+    setSearchedCity({ name: ev.target.value });
+  };
+
+  const filteredCities = searchedCity.name
+    ? cities.filter(city => city.name.includes(searchedCity.name))
+    : cities;
+  const citiesList = filteredCities.map(city => (
+    <a key={city.id} href="#">
+      {city.name}
+    </a>
+  ));
+  const dropdownContent = isOpen ? (
+    <div className={styles.dropdownContent}>
+      <input
+        className={styles.citySearch}
+        onChange={onSearch}
+        type="text"
+        placeholder="Search for a city"
+      />
+      {citiesList}
+    </div>
+  ) : null;
 
   return (
     <div className={styles.dropdown}>
-      <button id={styles.dropdownBtn} className={props.className} onClick={showCities}>Cities</button>
-      <div className={styles.dropdownContent}>{citiesList}</div>
+      <button className={props.navBtnClassName} onClick={showCities}>
+        Cities
+      </button>
+      {dropdownContent}
     </div>
   );
 }
--- a/src/components/Homepage/Nav/DropdownMenu.js
+++ b/src/components/Homepage/Nav/DropdownMenu.js
@@ -1,0 +1,35 @@
+import React, { useState } from 'react';
+
+export function DropdownMenu(props) {
+  const { styles } = props;
+
+  const [isOpen, setIsOpen] = useState(false);
+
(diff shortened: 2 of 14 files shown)


src/components/Homepage/Homepage.js
```diff
@@ -0,0 +1,37 @@
+import React, { useState, useEffect } from 'react';
+import * as homepageStyles from '../../styles/homepage.module.css';
+import { Nav } from './Nav/Nav.js';
+
+export function Homepage(props) {
```
virzhinia-petkova-redshelf: 1. it's much better not to export the function directly. Define it first and then export it at the bottom of the file. If you don't do that you'll have a hard time adding prop types. Read more about it here https://reactjs.org/docs/typechecking-with-proptypes.html#gatsby-focus-wrapper
2. The convention for naming react components is PascalCase aka HomePage

src/components/Homepage/Homepage.js
```diff
@@ -0,0 +1,37 @@
...
+  const [time, setTime] = useState('Loading...');
```
virzhinia-petkova-redshelf: If you'd like to have loading show up, use a different state instance. It's weird to change the data you have for your state with something unrelated - think of it as what is the value of time, not what the UI should render. You can use null for initial value here.

src/components/Homepage/Homepage.js
```diff
@@ -0,0 +1,37 @@
...
+  useEffect(() => {
+    const interval = setInterval(() => setTime(getCurrentTime()), 1000);
+    return () => clearInterval(interval);
+  }, []);
+
+  const getCurrentTime = () => {
+    const date = new Date();
+    const hours = date.getHours() < 10 ? `0${date.getHours()}` : date.getHours();
+    const minutes = date.getMinutes() < 10 ? `0${date.getMinutes()}` : date.getMinutes();
+    const seconds = date.getSeconds() < 10 ? `0${date.getSeconds()}` : date.getSeconds();
```
virzhinia-petkova-redshelf: all of those looks very similar - can you create a reusable function for them?

src/components/Homepage/Homepage.js
```diff
@@ -0,0 +1,37 @@
+import React, { useState, useEffect } from 'react';
+import * as homepageStyles from '../../styles/homepage.module.css';
```
virzhinia-petkova-redshelf: please use camelCase for such things aka homePage...

src/components/Homepage/Homepage.js
```diff
@@ -0,0 +1,37 @@
...
+    const minutes = date.getMinutes() < 10 ? `0${date.getMinutes()}` : date.getMinutes();
+    const seconds = date.getSeconds() < 10 ? `0${date.getSeconds()}` : date.getSeconds();
+    const currentTime = `${hours}:${minutes}:${seconds}`;
+
+    return currentTime;
+  };
+
+  return (
+    <>
+      <header>
```
virzhinia-petkova-redshelf: Is this page the only place you're going to use the header? What about the feedback/info page? Wouldn't you like to use it there as well? 

It will be a far better to use it as a separate component. The footer too. I'd suggest you think about structure/components and see if you can come up with a more robust structure.

src/components/Homepage/Homepage.js
```diff
@@ -0,0 +1,37 @@
...
+        <h1>Weather Forecast</h1>
```
virzhinia-petkova-redshelf: move this to a constant please

src/components/Homepage/Homepage.js
```diff
@@ -0,0 +1,37 @@
...
+      </header>
+      <section className="layout">
+        <Nav styles={homepageStyles} />
+        <main></main>
```
virzhinia-petkova-redshelf: why do you have an empty main? Also main should be the one the wraps all components in the main content(this does not include header/footer/etc), see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main

src/components/Homepage/Homepage.js
```diff
@@ -0,0 +1,37 @@
...
+      </section>
+      <footer>
+        <span>Current time: {time}</span>
```
virzhinia-petkova-redshelf: constant please

src/components/Homepage/Nav/DropdownMenu.js
```diff
@@ -0,0 +1,35 @@
+import React, { useState } from 'react';
+
+export function DropdownMenu(props) {
+  const { styles } = props;
+
+  const [isOpen, setIsOpen] = useState(false);
+
+  const showCities = ev => {
+    ev.preventDefault();
```
virzhinia-petkova-redshelf: you don't need to use preventDefault() since this is just a button

src/components/Homepage/Nav/DropdownMenu.js
```diff
@@ -0,0 +1,35 @@
...
+
+    const dropdownContent = [...document.getElementsByClassName(styles.dropdownContent)][0];
+    if (isOpen) {
+      setIsOpen(false);
+      dropdownContent.style.display = 'none';
```
virzhinia-petkova-redshelf: avoid altering css from js if you can help it

src/components/Homepage/Nav/DropdownMenu.js
```diff
@@ -0,0 +1,35 @@
...
+    } else {
+      setIsOpen(true);
+      dropdownContent.style.display = 'inline-block';
+    }
+  };
+
+  return (
+    <>
+      <div className={styles.dropdown}>
+        <button className={styles.dropdownBtn} onClick={showCities}>
```
virzhinia-petkova-redshelf: do you have an input field for filtering?

src/components/Homepage/Nav/DropdownMenu.js
```diff
@@ -0,0 +1,35 @@
...
+          Cities
+        </button>
+        <div className={styles.dropdownContent}>
+          <a href="#">Sofia</a>
+          <a href="#">Athens</a>
+          <a href="#">Istanbul</a>
+        </div>
```
virzhinia-petkova-redshelf: 1. Are those actual links that take you anywhere? If they're not, using a simple unordered list with list items inside will be a better fit.
2. Please create an array of objects that hold information for your cities(name, id) and render it instead of directly using them in the html

src/components/Homepage/Nav/Nav.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { DropdownMenu } from './DropdownMenu.js';
+
+export function Nav(props) {
+  return (
+    <>
+      <nav>
```
virzhinia-petkova-redshelf: not really sure nav is the right element to use here as you don't really use navigation in the dropdown

src/components/Homepage/Nav/Nav.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { DropdownMenu } from './DropdownMenu.js';
+
+export function Nav(props) {
+  return (
+    <>
```
virzhinia-petkova-redshelf: this is not needed, fragments are used when you have more than one element at the same level and need something to wrap than so react works

src/components/Homepage/Nav/Nav.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { DropdownMenu } from './DropdownMenu.js';
+
+export function Nav(props) {
```
virzhinia-petkova-redshelf: what is your idea with this component?

src/components/Homepage/Nav/Nav.js
```diff
@@ -0,0 +1,12 @@
...
+  return (
+    <>
+      <nav>
+        <DropdownMenu styles={props.styles} />
```
virzhinia-petkova-redshelf: it's better to directly import styles to the component that uses them rather than passing them down

src/styles/homepage.module.css
```diff
@@ -0,0 +1,82 @@
+header {
```
virzhinia-petkova-redshelf: please separate the styles for different components in different files, see https://create-react-app.dev/docs/adding-a-css-modules-stylesheet/

src/styles/homepage.module.css
```diff
@@ -0,0 +1,82 @@
...
+}
+
+nav {
+  display: flex;
+  border-style: solid;
+  border-width: 2px;
+}
+
+.dropdownBtn {
+  background-color: #3498db;
```
virzhinia-petkova-redshelf: please create variables for colors

src/styles/homepage.module.css
```diff
@@ -0,0 +1,82 @@
...
+.dropdown {
+  /* position: relative; */
+  display: flex;
+  flex-direction: column;
+  width: 10%;
+}
+
+.dropdownContent {
+  display: none;
+  /* position: absolute; */
```
virzhinia-petkova-redshelf: please remove commented code

src/components/Homepage/Nav/DropdownMenu.js
```diff
@@ -0,0 +1,35 @@
...
+      setIsOpen(false);
+      dropdownContent.style.display = 'none';
+    } else {
+      setIsOpen(true);
+      dropdownContent.style.display = 'inline-block';
+    }
+  };
+
+  return (
+    <>
```
virzhinia-petkova-redshelf: I don't think you need this wrapper

virzhinia-petkova-redshelf (commented):
Overall good job. I left a few comments. 

Also, please provide brief a description about what work was done when opening the PR.

src/components/Homepage/Homepage.js
StefanAtanasov98: Could you elaborate on this? I assume we want it to be constant so that we can manipulate it safely/easily in the future or..?

src/components/Homepage/Nav/Nav.js
StefanAtanasov98: My idea is that Nav contains more things (buttons I guess) than just DropdownMenu, so that's why I think it should be a parent component for DropdownMenu as well as the other buttons

src/components/Homepage/Nav/Nav.js
StefanAtanasov98: I imagined that when I click on a city which interests me in the dropdown, the view changes and on the page the user can see the weather for that specific city and have options like daily, weekly, monthly forecast and nothing else. From your comment I'm guessing that the view shouldn't change?

src/components/Homepage/Homepage.js
virzhinia-petkova-redshelf: You can refer to this conversation as it's discussing the same question - https://github.com/ximet/ReactTask/pull/290/files#r975016810

src/components/Homepage/Nav/Nav.js
virzhinia-petkova-redshelf: If you switch the view, you can leave it, it's fine. I like your idea. :)

src/components/Homepage/Nav/Nav.js
virzhinia-petkova-redshelf: what buttons for example?

StefanAtanasov98: Maybe the ones for the forecast options like "Daily", "Weekly", etc.

virzhinia-petkova-redshelf: okay

src/components/Homepage/Nav/DropdownMenu.js
```diff
@@ -0,0 +1,35 @@
+import React, { useState } from 'react';
+import * as styles from '../../../styles/DropdownMenu.module.css';
+
+const initialCitiesState = [
+  { name: 'Sofia', id: Math.random() },
+  { name: 'Athens', id: Math.random() },
+  { name: 'Istanbul', id: Math.random() }
+];
```
virzhinia-petkova-redshelf, line 8: you can just hard code something instead of using math.random - you'd need the id to be a constant value

src/components/Homepage/Nav/DropdownMenu.js
virzhinia-petkova-redshelf: Can you use a condition in JSX instead of adjusting the css? Something like(pseudo code):
`{isOpen ? dropdownContent : null}`

src/components/Homepage/Homepage.js
```diff
@@ -0,0 +1,19 @@
+import React, { useState, useEffect } from 'react';
+import * as styles from '../../styles/HomePage.module.css';
+import { Header } from '../Header/Header.js';
+import { Footer } from '../Footer/Footer.js';
+import { Nav } from './Nav/Nav.js';
+
+function HomePage(props) {
+  return (
+    <>
+      <Header />
```
virzhinia-petkova-redshelf: would you want to reuse the header and footer for other pages without adding the component to each page? Where will you put it to be able to do that?

src/styles/DropdownMenu.module.css
```diff
@@ -0,0 +1,28 @@
+:root {
```
virzhinia-petkova-redshelf: if there a way to declare the variables in a way you can use them everywhere?

virzhinia-petkova-redshelf (commented):
please provide a description & screenshot for your PR

src/components/Homepage/Homepage.js
StefanAtanasov98: Oh, right, of course. If we're gonna use the same header and footer, I should put the components in App.jsx

virzhinia-petkova-redshelf (approved):

PR #298, opened 2022-09-20
Modified Search and added LocationPage

ViliusGaizutis:
Screenshot 2022-09-20 170219

In the preview above you can see that app shows weather for New York. If you would like to see weather in Riga, for instance, you would type "riga" into search bar and click on "Riga, Latvia", and it would go to a different route, changing the URL address accordingly.

Modified search:
1. Added `RequestDataWrapper` component to `SearchResultList` component to better render request result (data, loading, error).
2. Added `Link` component to `SearchResultItem` component that leads to a specific location route (i.e. `/location/${id}`).
3. Added ability to automatically close search result modal on route change.

Added location page:
1. Modified webpack.dev.js to allow paths other than '/' and also nested paths.
2. Moved `Header` component to each page component to allow getting URL params.

Modified geolocation:
1. Used `useParams` hook to get `locationId`.


client/src/components/AboutPage/AboutPage.tsx
```diff
@@ -1,9 +1,15 @@
...
 
+// Components
+import Header from 'components/Header/Header';
+
 const AboutPage: FunctionComponent = () => (
-  <main>
-    <h1>About Page</h1>
-  </main>
+  <>
+    <Header />
```
JuliaShavyrova, line 8: please create layout component with using `<Outlet/>` to reduce number of `Header` component renders

JuliaShavyrova (approved):

PR #299, opened 2022-09-20
Add position in URL

AndrewHelix:
When app start, user can see weather in the current location
When user search a city or choose it in Country page, coordinates add to URL. Thereafter you can save page with selected location in bookmarks or send to a friend :)


src/components/main/Main.tsx
```diff
@@ -1,29 +1,43 @@
...
 
+  const { coordinates } = useParams();
+  const [paramLongitude, paramLatitude] = (coordinates || '').split(',');
+
   const {
     state: { position, positionError }
   } = useContext(positionContext);
 
-  async function fetchData(lon: number, lat: number) {
+  const { current: positionRef } = useRef<PositionRef>({ lon: 0, lat: 0 });
```
JuliaShavyrova: what is the reason to use ref instead of simple state here?

AndrewHelix: I don't see the reason to use state for this data. 
I get user location from context and selected position (new city form favorite or search) form URL. I'll get new data from URL on every render. I think I don't need useState for result info about position. 
I use useRef for save info in variable.

JuliaShavyrova: no need to use ref for it, use plain variable.

AndrewHelix: when I work with timerId you said me don't use plain  variable.
in which case I need to use ref, in which  plain  variable?

JuliaShavyrova: if not touching cases related to DOM manipulation, refs should be used only if you need variable inside your component, that should be stable between re-renders (e.g, another solutions, like plain variable and state, doesn't work). in case with timerId you mentioned i advised to use ref because: 
1. previously you had `let` variable outside of component body and in case you'd think in future about moving this functionality to custom hook and reuse in few places, it could cause bugs
2. just to try use ref in "classic" case where it can be useful

src/components/main/Main.tsx
```diff
@@ -36,17 +48,17 @@ const Main: FC = () => {
...
     if (selectDays) {
       getWeather<{ forecast: DailyWeatherType[] }>(
         '/forecast/daily/',
-        position.longitude,
-        position.latitude,
+        resultLongitude,
+        resultLatitude,
         { periods: selectDays, dataset: 'full' }
       ).then(res => setDailyWeather(res.forecast));
     }
```
JuliaShavyrova, line 54: why do you have deps array [resultLongitude, resultLatitude] in one case, but another one doesn't include these variables?

AndrewHelix: Do you mean useEffect with selectDays?
In this case I do request only when change selected days.

src/components/main/Main.tsx
```diff
@@ -1,29 +1,41 @@
...
-import { defaultCurrentWeather, defaultLocationInfo } from './defaultStates';
+import { defaultLocationInfo } from './defaultStates';
 import MainCard from './MainCard';
 import MainHourlyCard from './MainHourlyCard';
 import MainDailyCard from './MainDailyCard';
 
+type PositionRef = {
+  lon: number | string;
+  lat: number | string;
+};
```
JuliaShavyrova: type not used anywhere, please remove

src/components/main/Main.tsx
JuliaShavyrova: resultLongitude and resultLatitude should be mentioned in  second useEffect deps array

JuliaShavyrova: and actually, you can move this request to function  like `handleDaysSelected`, that will be placed instead of `onChange={e => setSelectDays(e.target.value)}`, because it's the only way to update `selectDays` state

JuliaShavyrova (approved):

PR #300, opened 2022-09-21
Feedback form

Auryte:
Here are the newest files for feedback form. All the minor changes and tweaks from the older files have made it here as well. Is it ok or should I make another branch for small corrections?


client/src/API/get.ts
```diff
@@ -8,8 +8,8 @@ export const createToken = async () => {
     const result = await axios.post(`http://localhost:5000/authorize/token?expire_hours=2`);
     window.localStorage.setItem('token', result.data.access_token);
     return result.data;
-  } catch (err) {
-    throw new Error((err as Error).message);
+  } catch (error) {
+    throw new Error((error as Error).message);
```
AndrewHelix, line 12: also you can use `if (error instanceof Error) {}`

client/src/contexts/FeedbackContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+  addFeedback: () => {
+    return null;
+  }
+});
+
+export const FeedbackProvider: FC<React.ReactNode> = ({ children }) => {
+  const [feedback, setFeedback] = useState<Feedback[]>([]);
+
+  const fetchFeedback = useCallback(() => {
+    const feedbackData = getFromStorage('feedback');
```
AndrewHelix: create const FEEDBACK_LS_LABEL

client/src/contexts/FeedbackContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+    if (feedbackData) {
+      setFeedback(feedbackData);
+    } else {
+      setInStorage('feedback', FeedbackData);
+      setFeedback(FeedbackData);
```
AndrewHelix: in both variants you use setFeedback. you can call this function after condition
`FeedbackData` and `feedbackData` look very similar

client/src/contexts/FeedbackContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+    }
+  }, [setFeedback]);
+
+  useEffect(() => {
+    fetchFeedback();
+  }, [fetchFeedback]);
+
+  const feedbackProvider = useMemo(() => {
```
AndrewHelix: I think it's not `Provider`. It is `Value `

client/src/contexts/FeedbackContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+    const addFeedback = (newFeedback: Feedback) => {
+      setInStorage('feedback', [newFeedback, ...feedback]);
```
AndrewHelix: you can do this logic `[newFeedback, ...feedback]` in LocalStorage API. And send in `setInStorage` only new feedback. 
else you need to get old feedback, change array in context and send to localStorage.

client/src/components/Data/FeedbackData.ts
```diff
@@ -0,0 +1,31 @@
...
+  {
+    id: '3',
+    nickname: 'Kris',
+    email: 'kris@mail.com',
+    title: 'works',
+    review:
+      'Lorem ipsum dolor sit amet consectetur adipisicing elit. consequuntur vel vitae commodi alias voluptatem est voluptatum ipsa quae.',
+    rating: 3
+  }
+];
```
AndrewHelix: why do you use hardcoded data? 
can you use real data only from localStorage?

client/src/API/localStorage.ts
```diff
@@ -0,0 +1,18 @@
+import { Feedback } from 'types';
+
+export const getFromStorage = (name: string): Feedback[] | null => {
+  const valueStr = localStorage.getItem(name);
+  if (valueStr === null) {
+    return null;
+  }
+  return JSON.parse(valueStr);
```
ViliusGaizutis: You can shorten the above to `return !valueStr ? null : JSON.parse(valueStr);`

client/src/components/Button/index.tsx
```diff
@@ -2,14 +2,15 @@ import React, { FC, ReactNode } from 'react';
...
+  ThemeBtn = 'themeBtn',
+  IconBtn = 'searchBtn',
+  FormBtn = 'formBtn'
 }
 interface ButtonProps {
   type: 'button' | 'reset' | 'submit' | undefined;
   children: ReactNode;
   className: string;
-  onClick?: () => void;
+  onClick?: (event: any) => void;
```
ViliusGaizutis: I think it would be better to define type, i.e. `(event: MouseEvent<HTMLButtonElement>)` .

client/src/reducers/feedbackReducer.ts
```diff
@@ -0,0 +1,35 @@
...
+};
+export const CREATE_FEEDBACK_SUCCESS = 'CREATE_FEEDBACK_SUCCESS';
+export const FEEDBACK_FORM_RESET = 'FEEDBACK_FORM_RESET';
+
+const feedbackReducer = (state: Feedback, { type, payload }: ActionData) => {
+  switch (type) {
+    case CREATE_FEEDBACK_SUCCESS:
+      return {
+        ...state,
+        [`${payload?.name}`]: payload?.data,
```
AndrewHelix, line 21: `[payload!.name]: payload!.data`

client/src/API/get.ts
Auryte: In this case eslint gives an error: Expected to return a value at the end of async arrow function.eslint[consistent-return]

client/src/components/Button/index.tsx
Auryte: Actually I have to write`  (event: FormEvent <HTMLButtonElement> )`

client/src/contexts/FeedbackContext.tsx
Auryte: Oh yes - you mean there shouldn't be any "magic strings" ? :)

client/src/components/Data/FeedbackData.ts
```diff
@@ -0,0 +1,31 @@
+import { Feedback } from 'types';
+
+export const FeedbackData: Feedback[] = [
```
JuliaShavyrova: please use camelCase for variables

client/src/contexts/FeedbackContext.tsx
Auryte: yes, it could be feedbackProviderValue

client/src/components/FeedbackForm/FeedbackForm.tsx
```diff
@@ -0,0 +1,94 @@
...
+    nickname: '',
+    email: '',
+    title: '',
+    review: '',
+    rating: 0
+  });
+
+  const handleChange = (event: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => {
+    const { name, value } = event.target;
+    dispatch({ type: CREATE_FEEDBACK_SUCCESS, payload: { name, data: value } });
```
JuliaShavyrova: type with name CREATE_FEEDBACK_SUCCESS for updating input values looks not very suitable, please rename

client/src/contexts/FeedbackContext.tsx
AndrewHelix: yes:)

client/src/components/Data/FeedbackData.ts
Auryte: When you or any other user opens this page, they see nothing except form inputs because their localStorage is empty. To understand better the design and how everything works, I made the mock data.

client/src/components/FeedbackForm/FeedbackForm.tsx
```diff
@@ -0,0 +1,94 @@
...
+          return (
+            <li key={`rating-${i + 1}`}>
+              <Input
+                type="radio"
+                name="rating"
+                id={`num${i + 1}`}
+                value={state.rating + i + 1}
+                onChange={handleChange}
+                label={<BsStar />}
+                checked={state.rating === i + 1}
```
JuliaShavyrova: i see `i + 1` too many times, please store it in variable

client/src/components/FeedbackForm/FeedbackForm.tsx
```diff
@@ -0,0 +1,94 @@
...
+        onChange={handleChange}
+        label="* Name"
+      />
+      <Input
+        type="email"
+        name="email"
+        id="emailInput"
+        value={state.email}
+        onChange={handleChange}
+        label="* Email"
```
JuliaShavyrova: you can create utility function to get required field label

client/src/components/Data/FeedbackData.ts
Auryte: it is easier to work, because you don't need to submit form many times

client/src/components/FeedbackForm/FeedbackForm.tsx
```diff
@@ -0,0 +1,94 @@
...
+        id="titleInput"
+        value={state.title}
+        onChange={handleChange}
+        label="* Review Title"
+      />
+      <Input
+        type="textarea"
+        name="review"
+        id="reviewInput"
+        value={state.review}
```
JuliaShavyrova: state can be destructured

client/src/components/FeedbackItem/FeedbackItem.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { BsStar, BsStarFill } from 'react-icons/bs';
+
+type ItemProps = {
+  item: Feedback;
+};
+
+const FeedbackItem: FC<ItemProps> = ({ item }) => {
+  const starArray = [...Array(5).keys()].map(i => {
+    return i + 1;
+  });
```
JuliaShavyrova: you can create separate controlled input for star rating and reuse it both in feedback form and here

client/src/components/Input/Input.module.scss
```diff
@@ -0,0 +1,19 @@
...
+        color: white;
+        font-size: 1rem;
+        font-weight: 400;
+      
+        &::placeholder {
+          color: white;
+        }
+      }
+      
+}
```
JuliaShavyrova: npm run format?

client/src/components/Data/FeedbackData.ts
AndrewHelix: You can add some feedbacks in LS. When you open this page, you can get data from LS. 
I think in real case app will use database and all users will see feedbacks from DB

client/src/components/Input/Input.tsx
```diff
@@ -0,0 +1,35 @@
+import React, { FC, ReactNode } from 'react';
+import styles from './input.module.scss';
+
+type InputProps = {
```
JuliaShavyrova, line 5: maybe you can use built-in `InputHTMLAttributes` type

client/src/components/Input/Input.tsx
```diff
@@ -0,0 +1,35 @@
...
+        {type === 'textarea' ? (
+          <textarea name={name} id={id} value={value} onChange={onChange} />
+        ) : (
+          <input
+            type={type}
+            name={name}
+            id={id}
+            value={value}
+            onChange={onChange}
+            //   data-validators="required, string, min-length(7), max-length(15)"
```
JuliaShavyrova: commented code :)

client/src/reducers/feedbackReducer.ts
Auryte: Thanks, I didn't know about this option :)

Auryte: eslint marks as warning - Forbidden non-null assertion.eslint[@typescript-eslint/no-non-null-assertion]

client/src/contexts/FeedbackContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+    };
+    return { feedback, setFeedback, addFeedback };
+  }, [feedback]);
+
+  return <FeedbackContext.Provider value={feedbackProvider}>{children}</FeedbackContext.Provider>;
+};
+
+export default FeedbackContext;
```
JuliaShavyrova, line 45: 1. please remove `setFeedback` from deps array, it's not needed;
2.  `setFeedback` not needed to be passed as context value. the only usage i found looks this way:
```
setFeedback([state, ...feedback]);
addFeedback(state);
```
and setting feedback code can be moved inside `addFeedback` function

client/src/reducers/feedbackReducer.ts
```diff
@@ -0,0 +1,35 @@
+import { Feedback } from 'types';
+import { uid } from 'utils/uid';
+
+export type ActionData = {
+  type: string;
+  payload?: {
+    name: string;
+    data?: string | number | readonly string[] | undefined;
```
JuliaShavyrova: why `readonly`?

client/src/contexts/FeedbackContext.tsx
JuliaShavyrova: cannot agree with point about customizing `setInStorage`, because it is defined as shared function and should be workable for different cases (for example, it can be used for another component with different logic of setting array in local storage)

client/src/components/FeedbackForm/FeedbackForm.tsx
Auryte: Could you explain what do you mean?

JuliaShavyrova: for every required field you need to write label like this: '* <field name>'. you can create utility function that will add star symbol for field name instead of adding it to every label manually

client/src/components/Input/Input.tsx
Auryte: Sorry :)

client/src/reducers/feedbackReducer.ts
Auryte: something what eslint or ts suggested in vscode :) ... I remove it

client/src/contexts/FeedbackContext.tsx
```diff
@@ -0,0 +1,46 @@
...
+  }
+});
+
+const LOCALSTORAGE_LABEL = 'feedback';
+
+export const FeedbackProvider: FC<React.ReactNode> = ({ children }) => {
+  const [feedback, setFeedback] = useState<Feedback[]>([]);
+
+  const fetchFeedback = useCallback(() => {
+    const feedbackData = getFromStorage(LOCALSTORAGE_LABEL);
```
AndrewHelix: I can't understand what in this variable `LOCALSTORAGE_LABEL`
this name does not mean that it because it is common name. You can have multiple variables and name each `LOCALSTORAGE_LABEL`
I think the best name will be `FEEDBACK_LOCALSTORAGE_LABEL`

client/src/components/FeedbackItem/FeedbackItem.tsx
```diff
@@ -1,22 +1,24 @@
...
-      {starArray.map(i => {
-        return item.rating >= i ? <BsStarFill key={i} /> : <BsStar key={i} />;
-      })}
-      <h3>{item.title}</h3>
+      <InputRating
+        type="hidden"
+        name="postRating"
+        id="postNum"
+        value={item.rating}
+        onChange={() => {
```
JuliaShavyrova: you can mark onChang prop as optional

client/src/components/FeedbackStats/FeedbackStats.tsx
```diff
@@ -1,11 +1,31 @@
...
+import { Feedback } from 'types';
 
 const FeedbackStats = () => {
   const { feedback } = useContext(FeedbackContext);
+  const feedbackRatingAvg = feedback.reduce(
+    (prev: number, current: Feedback, index: number, array: Feedback[]): number => {
+      const sum = prev + +current.rating / array.length;
+      return sum;
+    },
+    0
```
JuliaShavyrova: please move this code to separate function

client/src/contexts/FeedbackContext.tsx
```diff
@@ -1,49 +1,46 @@
...
-  }, [feedback]);
-
-  return <FeedbackContext.Provider value={feedbackProvider}>{children}</FeedbackContext.Provider>;
+  const addFeedback = (newFeedback: Feedback) => {
+    setInStorage(LOCALSTORAGE_LABEL, [newFeedback, ...feedback]);
+    setFeedback([newFeedback, ...feedback]);
+  };
+  const feedbackProviderValue = () => {
+    return { feedback, addFeedback };
+  };
```
JuliaShavyrova: why do you need function here?

client/src/components/FeedbackItem/FeedbackItem.tsx
Auryte: Yes, I have done that, but then I got this error in the console
Screenshot 2022-09-23 131435

client/src/contexts/FeedbackContext.tsx
Auryte: To avoid useMemo :)

Auryte: I get this error : The 'feedbackProviderValue' object (at line 37) passed as the value prop to the Context provider (at line 40) changes every render. To fix this consider wrapping it in a useMemo hook.eslint[react/jsx-no-constructed-context-values]

JuliaShavyrova: ok, you are trying to avoid error. please explain me what way this function will be better than plain variable from any perspective?

Auryte: Nothing better, I just get eslint error.
Screenshot 2022-09-23 141632

Auryte: what would you suggest to do?

JuliaShavyrova: eslint error is not a reason, i can add to your config such a rules that will show errors for every code line. you are a  code owner, please explain differences between these two solutions as react developer

client/src/components/FeedbackItem/FeedbackItem.tsx
JuliaShavyrova: try to use disabled attribute for read-only input, in this case you won't have this error when no onChange provided

client/src/contexts/FeedbackContext.tsx
Auryte: The first one declares function while the second one declares a simple object. In JavaScript, functions are also objects, so there is almost no difference between the two except that in the first example you can call as a function. I think, those two approaches are pretty much exactly the same. A function instance is just an object . But, generally this is not how a function is used. A function is a procedure to be performed to achieve something. So I would use the second example and set to off eslint  ‘react/jsx-no-constructed-context-values’.

client/src/components/FeedbackItem/FeedbackItem.tsx
Auryte: No, it is the same result with disabled. But I removed checked.

client/src/contexts/FeedbackContext.tsx
JuliaShavyrova: as far as you say that these two solutions have no difference and eslint doesn't like one of these solutions - that means second one also won't be the best choice according to exactly this eslint rule. and all that means you have 2 options here:
1. remove this rule and fix your code
2. fix your code according to eslint rule

Auryte:
Screenshot 2022-09-27 174901
Screenshot 2022-09-27 174942
Screenshot 2022-09-27 175124
Screenshot 2022-09-27 175645

JuliaShavyrova:
your last commit contains too much new code (not requested changes only) and now PR doesn't meet recommended size we defined before (approximately 25 files). please divide your code and create a few small PRs instead of one large

Auryte:
So it is exactly 26. Isn't it approximately? :)

JuliaShavyrova:
26 files you pushed with 1 last commit, 40 files changed overall in PR

Auryte:
I want to show changes when I am more or less satisfied with the result and it is worth showing. Should I show something half working or looking ugly? And what should I do now? Create new branch for validation and new for design files?

JuliaShavyrova:
every big feature can be divided to small parts and every PR should contain code that produces no compilation errors.
so in your case you can divide current code as you suggested by files (design and functionality) or, in case you don't want to separate styles from components (and i'd recommend to follow this way), you can divide code by smaller features (form, rating, feedback list, etc)

Auryte:
If I were to show an unfinished result, I would do all the time fixing comments and would never finish anything :) How about a real project? Should I do each task in small steps and wait for comments? But comments would be - here and there needs to be finished, this or that isn't working and etc. ? Please, explain me, because I feel I don't understand something.

Auryte:
Maybe I split the code too much, resulting in too many files? Maybe I should make bigger components?

JuliaShavyrova:
1. on real projects you can meet the same requirements about PR recommended size, it's not my invention
2. you can divide big feature to smaller ones, work on them step by step and create PR every time you finished one of these small features
3. while your PR is reviewed, you can create another branch and work with another feature
4. when you have comments in PR, fix them in current PR, but don't add non-related new code
5. about bigger components - it's not the option, big components are hard to read and support in future

PR #305, opened 2022-09-28
Added AboutPage

ViliusGaizutis:
Screenshot 2022-09-28 161358
Screenshot 2022-09-28 161451
Screenshot 2022-09-28 162801
Preview of about page.

Revisions:
1. [Tweaked styles, defined const HEADER_HEIGHT](https://github.com/ViliusGaizutis/ReactTask/commit/6a429d9d9a753437ee2d61c23a0a89824ccfb8e1)
2. [Added images and image type definitions](https://github.com/ViliusGaizutis/ReactTask/commit/ff0d1e3316f88518ddb8e08fae45b04489099b98)
3. [Added AboutPage](https://github.com/ViliusGaizutis/ReactTask/commit/3596836975b3dadbb7879091901f20068077e874)


client/src/assets/images/svg/index.tsx
```diff
@@ -55,3 +57,9 @@ export const IconMenu = () => (
...
   </svg>
 );
+
+export const Wave = ({ reversed }: { reversed: boolean }) => (
```
JuliaShavyrova: please create interface for `{ reversed: boolean }` and add type definitions for components/functions

client/src/components/AboutPage/AboutHeroSection/AboutHeroSection.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { Container, Flex } from 'styles/global';
+import * as S from '../AboutPage.styles';
+
+interface AboutHeroSectionProps {
+  scrollToRef: MutableRefObject<HTMLElement | null>;
+}
+
+const AboutHeroSection: FunctionComponent<AboutHeroSectionProps> = ({ scrollToRef }) => (
+  <S.AboutHeroSection>
+    <Wave reversed={false} />
```
JuliaShavyrova: if you mark `reserved` prop as optional and provide default value, it won't be needed to pass `reversed={false}` everywhere

client/src/components/AboutPage/AboutHeroSection/AboutHeroSection.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+// Styles
+import { Container, Flex } from 'styles/global';
+import * as S from '../AboutPage.styles';
+
+interface AboutHeroSectionProps {
+  scrollToRef: MutableRefObject<HTMLElement | null>;
+}
+
+const AboutHeroSection: FunctionComponent<AboutHeroSectionProps> = ({ scrollToRef }) => (
```
JuliaShavyrova: just interesting - why 'Hero'?

client/src/components/AboutPage/AboutHowSection/AboutHowSection.tsx
```diff
@@ -0,0 +1,59 @@
...
+interface AboutHowSectionProps {
+  customRef: RefObject<HTMLElement>;
+}
+
+const AboutHowSection: FunctionComponent<AboutHowSectionProps> = ({ customRef }) => (
+  <S.AboutHowSection id="how-it-works" ref={customRef}>
+    <Container>
+      <Flex directionColumn>
+        <h2>How It Works</h2>
+        {articles.map((article, i) => (
```
JuliaShavyrova: `article` can be destructured

client/src/components/AboutPage/AboutHeroSection/AboutHeroSection.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <S.AboutHeroContent>
+          <S.AboutHeadline>
+            <span>Weather App</span>
+            Find your weather predictions in your city
+          </S.AboutHeadline>
+          <p>Easy steps to predict the weather and make your day easier</p>
+          <Link to="/">
+            <Button>Get Started</Button>
+          </Link>
+          <Link to="#how-it-works" onClick={() => handleScrollTo(scrollToRef)}>
```
JuliaShavyrova: have you tried css first? 
https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior

client/src/components/AboutPage/AboutHowSection/AboutHowSection.tsx
```diff
@@ -0,0 +1,59 @@
...
+    content: 'Switch to light or dark mode and vice versa depending on your preference.',
+    image: { src: imgThemeSwitch, alt: 'Example of using theme-switch' }
+  }
+];
+
+interface AboutHowSectionProps {
+  customRef: RefObject<HTMLElement>;
+}
+
+const AboutHowSection: FunctionComponent<AboutHowSectionProps> = ({ customRef }) => (
```
JuliaShavyrova: please rename `customRef`, not clear why it's custom and why it's needed according to current name

client/src/components/AboutPage/AboutHeroSection/AboutHeroSection.tsx
ViliusGaizutis: "A hero section is a full screen section typically consisting of a background image, or video, or illustrations, or animations, with text and sometimes a call to action." (https://mbcreative.ca/blog/15-hero-section-design-inspiration/). Just from a random site from Google.

client/src/components/AboutPage/AboutHowSection/AboutHowSection.tsx
ViliusGaizutis: At first, I named it as just `ref`, but got the following warnings in the console: "Warning: AboutHowSection: `ref` is not a prop. Trying to access it will result in `undefined` being returned. If you need to access the same value within the child component, you should pass it as a different prop. (https://reactjs.org/link/special-props)" and "react-dom.development.js:67 Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?"

Should I just write `aboutHowSectionRef={aboutHowSectionRef}`?

client/src/components/AboutPage/AboutHeroSection/AboutHeroSection.tsx
ViliusGaizutis: I did now. I am not sure where to put it, though. In a `body`, an `AboutHowSection` styles? Tried both, but it does not work. Do I have to do something else? Also, I am not sure if this should work at all. I tried looking into how to scroll or jump to a particular section in a page with an `id` by clicking on `Link`, but haven't found anything. The only solution I managed to found was to use `element.scrollIntoView()`.

JuliaShavyrova: i've took a look at your code to check smooth-scrolling, so can give you few hints:
1. please use plain `a` tag for anchor links, no `Link` component is required, because it's not a react-router navigating
2. take a look on example in docs, it uses not only `scroll-behavior` (it should be added for container you want to be scrolled), but also defines `height` for container (100% value won't work, but 100vh will - maybe you can explain me why?) and uses `overflow` property

also few minor comments and questions:
1. please fix active state for sidebar
image

2. why have you moved `Header` out of `Layout` component?

client/src/components/AboutPage/AboutHowSection/AboutHowSection.tsx
JuliaShavyrova: if you won't use css, `aboutHowSectionRef` looks better than `customRef`

client/src/components/AboutPage/AboutHeroSection/AboutHeroSection.tsx
ViliusGaizutis: I have avoided creating yet another `div` for `Layout` just to set this smooth-scroll to work. What I have done is placed the required css in `<GlobalStyle/> ==> body ==>`
```
 #app {
      scroll-behavior: smooth;
      height: 100vh;
      overflow-x: hidden;
    }
```
Also I've tried looking into why this works only with `vh` units and found that "vh units will be sized to the "largest possible viewport". This means 100vh will be larger than the visible height when the URL bar is shown. While if its height is 100% it will always fill exactly the visible height, whether or not the URL bar is shown." (https://developer.chrome.com/blog/url-bar-resizing/). I'm not sure if this explains it, though.

Also I don't understand how you got sidebar to show both `li` as active. For me it is working fine. Though, I seem to forgot to set `li` named Dashboard as active on `/locations/:locationId` route.

JuliaShavyrova: "Also I don't understand how you got sidebar to show both li as active. For me it is working fine. Though, I seem to forgot to set li named Dashboard as active on /locations/:locationId route."
i've just navigated to about page :)

ViliusGaizutis: I still don't understand. It is working for me correctly.

JuliaShavyrova: dashboard navlink has '/' url, about - '/about'. looks like you need additional check for dashboard navlink active state
image

ViliusGaizutis: Strange. But why I don't get this result?

When I console.log `isActive` for links
```
  { to: '/', name: 'Dashboard' },
  { to: '/about', name: 'About' },
  { to: '/feedback', name: 'Feeback' }
```
I get 
Screenshot 2022-09-29 160448
Screenshot 2022-09-29 160647

JuliaShavyrova: skip it for now

JuliaShavyrova (approved):

PR #306, opened 2022-09-29
Feedback form2

Auryte:


client/src/components/Header/index.tsx
```diff
@@ -37,7 +37,7 @@ const Header: React.FC = ({ children }) => {
       </nav>
       <div className={styles.searchBox}>
         <div className={styles.search}>{children}</div>
-        <Button type="button" className="themeBtn" onClick={onThemeBtnClick}>
+        <Button type="button" className="ThemeBtn" onClick={onThemeBtnClick}>
```
JuliaShavyrova: if you are using enum it should be written like this: `className={ButtonStyles.ThemeBtn}`

client/src/components/Header/styles.module.scss
```diff
@@ -61,4 +61,20 @@
...
+      color: $footergrey;
+    }
+  }
+
+  .iconDark {
+    color: $lightgrey;
+
+    &:hover {
+      color: $footergrey;
+    }
```
JuliaShavyrova: please remove duplicate

client/src/components/InputRating/InputRating.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+const InputRating: FC<InputHTMLAttributes<HTMLInputElement> & InputRatingProps> = ({
+  type,
+  name,
+  id,
+  value,
+  onChange
+}) => {
+  return (
+    <div className={type !== 'hidden' ? styles.ratingContainer : styles.ratingContainerMsg}>
```
JuliaShavyrova: `type !== 'hidden'` should be defined as variable

client/src/components/InputRating/InputRating.tsx
```diff
@@ -0,0 +1,47 @@
...
+      <div className={styles.textContainer}>
+        <h3>{type !== 'hidden' ? '* Rating' : ''}</h3>
+        <div>{type !== 'hidden' && 'Fields marked with a * are required.'}</div>
+      </div>
```
JuliaShavyrova: do you need to render h3, div tags and `textContainer` if type is hidden? if no, please update the code

client/src/contexts/FeedbackContext.tsx
```diff
@@ -0,0 +1,45 @@
...
+    if (!feedbackData) setInStorage(FEEDBACK_LOCALSTORAGE_LABEL, mockFeedbackData);
+  }, []);
+
+  useEffect(() => {
+    fetchFeedback();
+  }, [fetchFeedback]);
+
+  const addFeedback = (newFeedback: Feedback) => {
+    setInStorage(FEEDBACK_LOCALSTORAGE_LABEL, [newFeedback, ...feedback]);
+    setFeedback([newFeedback, ...feedback]);
```
JuliaShavyrova: please define `[newFeedback, ...feedback]` as variable

client/src/pages/Feedback/index.tsx
```diff
@@ -1,10 +1,15 @@
...
 import SecondaryPagesLayout from 'components/layouts/SecondaryPagesLayout';
+import { FeedbackProvider } from 'contexts/FeedbackContext';
 import React from 'react';
 
 const Feedback = () => {
   return (
     <SecondaryPagesLayout>
-      <h1>Feedback Page</h1>
+      <FeedbackProvider>
+        <h2>Leave your Feedback</h2>
```
JuliaShavyrova: title definitely doesn't need context values, please move it out of provider

client/src/utils/stringCorrections.ts
```diff
@@ -1,11 +1,21 @@
...
+export const formatNameForUrl = (name: string): string => {
   return name.replace(/\s+/g, '');
 };
 
-export const resetLocationName = name => {
+export const separateCapitalizeName = (name: string): string => {
   return name.replace(/([a-z])([A-Z])/g, '$1 $2');
 };
+export const capitalizeFirstLetter = (name: string): string => {
+  return name.substring(0, 1).toUpperCase() + name.substring(1).replace(/([a-z])([A-Z])/g, '$1 $2');
```
JuliaShavyrova: `.replace(/([a-z])([A-Z])/g, '$1 $2')` was used in `separateCapitalizeName` function, so please reuse it here

client/src/components/InputRating/InputRating.tsx
```diff
@@ -15,11 +17,17 @@ const InputRating: FC<InputHTMLAttributes<HTMLInputElement> & InputRatingProps>
   onChange
 }) => {
   return (
-    <div className={type !== 'hidden' ? styles.ratingContainer : styles.ratingContainerMsg}>
-      <div className={styles.textContainer}>
-        <h3>{type !== 'hidden' ? '* Rating' : ''}</h3>
-        <div>{type !== 'hidden' && 'Fields marked with a * are required.'}</div>
-      </div>
+    <div
+      className={type !== INPUT_TYPE_HIDDEN ? styles.ratingContainer : styles.ratingContainerMsg}
```
JuliaShavyrova: i meant to move to variable the whole expression `type !== 'hidden'`, not just word 'hidden'

client/src/components/InputRating/InputRating.tsx
```diff
@@ -15,11 +17,17 @@ const InputRating: FC<InputHTMLAttributes<HTMLInputElement> & InputRatingProps>
...
+    >
+      {type !== INPUT_TYPE_HIDDEN ? (
+        <div className={styles.textContainer}>
+          <h3>* Rating</h3>
+          <div>Fields marked with a * are required.</div>
+        </div>
+      ) : (
+        ''
```
JuliaShavyrova: use `type !== INPUT_TYPE_HIDDEN && ...` instead of ternary operator, you don't need second part of condition

client/src/utils/stringCorrections.ts
```diff
@@ -10,7 +10,8 @@ export const separateCapitalizeName = (name: string): string => {
   return name.replace(/([a-z])([A-Z])/g, '$1 $2');
 };
 export const capitalizeFirstLetter = (name: string): string => {
-  return name.substring(0, 1).toUpperCase() + name.substring(1).replace(/([a-z])([A-Z])/g, '$1 $2');
+  const capitalizedName = name.substring(0, 1).toUpperCase() + name.substring(1);
+  return separateCapitalizeName(capitalizedName);
```
JuliaShavyrova: in previous version you had `.replace` for `name.substring(1)` only, but now it applies to the whole expression.
are you sure current code works as you expected?

Auryte: I am not sure, because, I don't have the other part of the code. I can not check or see the result. That is the downside of making separate branches for related components.  I'll fix it when I merge the next branch.

JuliaShavyrova: can't you test your own code with some examples directly inside app using console.log?
or, if you want to test it on real data, there is no big deal on switching to another branch and making the same small change for testing.
or at least you can read the code and analyze it

Auryte: Yes, it's working, but I will add it later, because in this branch it is not needed.

client/src/components/InputRating/InputRating.tsx
```diff
@@ -2,7 +2,7 @@ import React, { ChangeEvent, FC, InputHTMLAttributes } from 'react';
 import { BsStar, BsStarFill } from 'react-icons/bs';
 import styles from './InputRating.module.scss';
 
-const INPUT_TYPE_HIDDEN = 'hidden';
+const checkIfTypeNotHidden = 'type !== "hidden"';
```
JuliaShavyrova: Did you make a string with expression inside?

client/src/components/InputRating/InputRating.tsx
JuliaShavyrova: Looks like you reverted changes i talked about on previous review :)

JuliaShavyrova: Also you could use function you already created instead of typing manually “* Rating” but it can be fixed in next PR

JuliaShavyrova: Also you could use function you already created instead of typing manually “* Rating” but it can be fixed in next PR

JuliaShavyrova (approved):

PR #307, opened 2022-09-29
Form validation

Auryte:


JuliaShavyrova:
i'll check this PR after #306 will be merged because current one contains the same code

client/src/components/FeedbackForm/FeedbackForm.tsx
```diff
@@ -39,24 +52,50 @@ const FeedbackForm: FC = () => {
...
         name="reviewTitle"
         id="title"
         value={reviewTitle}
         onChange={handleChange}
       />
+      <div className={styles.error}>{reviewTitleError || ''}</div>
       <Input type="textarea" name="review" id="review" value={review} onChange={handleChange} />
-      <Button type="submit" className={ButtonStyles.FormBtn} onClick={handleSubmit}>
+      <div className={styles.error}>{reviewError || ''}</div>
+      <Button type="submit" className="FormBtn" onClick={handleSubmit}>
```
JuliaShavyrova: `className="FormBtn"` should it be a enum value?

client/src/constants/errors.ts
```diff
@@ -0,0 +1,22 @@
+import { capitalizeFirstLetter } from 'utils/stringCorrections';
+
+export const ERRORS = {
+  required: (field: string) => {
+    return `This ${capitalizeFirstLetter(field)} field is required!`;
+  },
+  'required-rating': (field: string) => {
```
JuliaShavyrova: why not camelCase?

client/src/constants/validatorStrings.ts
```diff
@@ -0,0 +1,7 @@
+export const VALIDATORS_STRINGS = {
+  rating: 'required-rating',
+  nickname: 'required, string, min-length(3), max-length(15)',
+  email: 'required, validate-email',
+  reviewTitle: 'required, min-length(4), max-length(25)',
+  review: 'required, min-length(7), max-length(250)'
```
JuliaShavyrova: defining rules as strings can cause bugs because of typo errors (for example someone removed space symbol from string or written word incorrectly). would be better to pass these rules as string array (and in perfect case it will be not a plain string, but enum/const value). i've already said the same to Andrew, so you can cooperate with him :)

client/src/components/Header/styles.module.scss
```diff
@@ -61,4 +61,4 @@
       color: $footergrey;
     }
   }
-}
+
```
JuliaShavyrova: curly bracket definitely should be returned, otherwise your app won't be built

JuliaShavyrova: also `.header` selector contains 2 definitions of `position` property, please remove one

client/src/components/FeedbackForm/FeedbackForm.tsx
```diff
@@ -39,24 +52,50 @@ const FeedbackForm: FC = () => {
     event: React.FormEvent<HTMLFormElement | HTMLTextAreaElement | HTMLButtonElement>
   ) => {
     event.preventDefault();
-    addFeedback(state);
-    dispatch({ type: FEEDBACK_FORM_RESET });
+    const inputsStateArray = [{ rating }, { nickname }, { email }, { reviewTitle }, { review }];
+    inputsStateArray.forEach((stateObj: object) => {
+      dispatch({
+        type: UPDATE_INPUT_VALUES,
+        payload: { name: getName(stateObj), data: getValue(stateObj) }
```
JuliaShavyrova: too complex function. you can write the same transformation much shorter using object like this `{ rating, nickname, email, reviewTitle, review }` and `Object.keys` method

client/src/constants/errors.ts
Auryte: I need to capitalize reviewTitle  and then to separate it - I need the result "Review Title". Or do you mean to change name of capitalizeFirstLetter() into camelCase() ?

JuliaShavyrova: No, I mean writing `requiredRating` instead of “required-rating”

client/src/constants/validator.ts
```diff
@@ -0,0 +1,22 @@
+import { validateEmail, validateName } from 'utils/validateRegex';
+
+export const VALIDATOR = {
+  required: (value: string) => {
+    return !!(value.length === 0);
```
JuliaShavyrova: `!value.length` is enough

client/src/constants/validator.ts
```diff
@@ -0,0 +1,22 @@
...
+  },
+  requiredRating: value => {
+    return !!(+value === 0);
+  },
+  string: (value: string) => {
+    return !validateName.test(value);
+  },
+  minLength: (value: string, param: number) => {
+    return !!(value.length < param);
```
JuliaShavyrova: ` < ` operator always returns boolean, no need to add `!!`

client/src/constants/validatorStrings.ts
```diff
@@ -0,0 +1,36 @@
...
+  MAX_LENGTH = 'maxLength',
+  VALIDATE_EMAIL = 'validateEmail'
+}
+
+const nameMinValue = '3';
+const nameMaxValue = '15';
+const titleMinValue = '4';
+const titleMaxValue = '25';
+const reviewMinValue = '7';
+const reviewMaxValue = '250';
```
JuliaShavyrova: why not numbers?

client/src/utils/validator.ts
```diff
@@ -0,0 +1,37 @@
+import { ERRORS } from '../constants/errors';
+import { VALIDATORS_STRINGS } from '../constants/validatorStrings';
+import { VALIDATOR } from '../constants/validator';
+
+interface validatorsArray {
```
JuliaShavyrova: type definitions should be in PascalCase, not camelCase

client/src/constants/validator.ts
```diff
@@ -0,0 +1,22 @@
...
+  },
+  maxLength: (value: string, param: number) => {
+    return !!(value.length > param);
+  },
+  validateEmail: (value: string) => {
```
JuliaShavyrova: you can use enum values here instead of prop name: `validateEmail` => `[Validations.VALIDATE_EMAIL]`

client/src/constants/validator.ts
```diff
@@ -0,0 +1,22 @@
+import { validateEmail, validateName } from 'utils/validateRegex';
+
+export const VALIDATOR = {
```
JuliaShavyrova: type definition missed

client/src/constants/validatorStrings.ts
```diff
@@ -0,0 +1,36 @@
...
+  email: [Validations.REQUIRED, Validations.VALIDATE_EMAIL],
+  reviewTitle: [
+    Validations.REQUIRED,
+    `${Validations.MIN_LENGTH}(${titleMinValue})`,
+    `${Validations.MAX_LENGTH}(${titleMaxValue})`
+  ],
+  review: [
+    Validations.REQUIRED,
+    `${Validations.MIN_LENGTH}(${reviewMinValue})`,
+    `${Validations.MAX_LENGTH}(${reviewMaxValue})`
```
JuliaShavyrova: instead of generating complex string and then parsing it, you can pass object with validator name and additional param

client/src/utils/validator.ts
```diff
@@ -0,0 +1,37 @@
...
+    const valName: string = validator.split(separator)[0];
+    const regExp = /[()]/;
+    const param: string = validator.split(regExp)[1];
+    validatorsNew.push({
+      valName,
+      param
+    });
+    return null;
+  });
+  return validatorsNew;
```
JuliaShavyrova: `map` method is used to create new array with the same length, based on array is called with. in your case this method is used incorrectly: you should either return new item (you're returning null) or use `forEach` method instead of `map`, because `forEach` doesn't require `return` statement. i'd recommend to use `map`, but use it in correct way

client/src/utils/validator.ts
```diff
@@ -0,0 +1,37 @@
+import { ERRORS } from '../constants/errors';
+import { VALIDATORS_STRINGS } from '../constants/validatorStrings';
+import { VALIDATOR } from '../constants/validator';
+
+interface validatorsArray {
+  valName: string;
+  param: string;
+}
+const createInputValidators = (name: string) => {
```
JuliaShavyrova: please add type definition for returned value

client/src/utils/validator.ts
```diff
@@ -0,0 +1,37 @@
...
+    });
+    return null;
+  });
+  return validatorsNew;
+};
+
+export const isInputValid = (name: string, value: string | number): string => {
+  const inputMessage: string[] = [];
+  const inputValidators = createInputValidators(name);
+  if (inputValidators) {
```
JuliaShavyrova: no need to check, `createInputValidators` always returns array

client/src/constants/validatorStrings.ts
Auryte: yeah I just thought the same :)

client/src/constants/validator.ts
Auryte: type for VALIDATOR and for ERRORS?

JuliaShavyrova: for every variable, that doesn't have type :)

client/src/types.ts
```diff
@@ -91,3 +91,21 @@ export interface Feedback {
...
+};
+
+export type Errors = {
+  required: (value: string) => string;
+  requiredRating: (value: string) => string;
+  string: (value: string) => string;
+  minLength: (value: string, param: string) => string;
+  maxLength: (value: string, param: string) => string;
+  validateEmail: (value: string) => string;
+};
```
JuliaShavyrova: both types can be defined more elegant way like this:
export type Validator = Record<Validations, (value: string, param?: number) => boolean>

client/src/utils/validator.ts
```diff
@@ -2,36 +2,19 @@ import { ERRORS } from '../constants/errors';
 import { VALIDATORS_STRINGS } from '../constants/validatorStrings';
 import { VALIDATOR } from '../constants/validator';
 
-interface validatorsArray {
+interface ValidatorsArray {
```
JuliaShavyrova: please think a bit more on interface name. your current interface is not an array, it's an object, so name cannot include 'array' word (exception - 'ArrayItem' or something like that)

JuliaShavyrova (commented):
good enough, few small suggestions left

JuliaShavyrova (approved):

PR #316, opened 2022-10-03
Added FeedbackPage, new Input variations (including StarRating)

ViliusGaizutis:
Screenshot 2022-10-03 124531
Screenshot 2022-10-03 124612
Preview of feedback page.

Revisions:
1. [Added FeedbackPage, created global Headline styled component](https://github.com/ViliusGaizutis/ReactTask/commit/8d00e714d5e7e372222618c0b28a9eef5d933f18)
2. [Added FeedbackFormSection and expanded Input variantions](https://github.com/ViliusGaizutis/ReactTask/commit/902f451504ae4c8e316125e621e7163c42fba27c)
3. [Added StarRating, created some styles mixins and global types](https://github.com/ViliusGaizutis/ReactTask/commit/2bada716f1d8d75f65daa120b97591f6e403debd)
4. [Removed unnecessary interface props](https://github.com/ViliusGaizutis/ReactTask/commit/0043265cc1a94ee532f947354d9ccb9d66492845)


client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx
```diff
@@ -0,0 +1,124 @@
...
+    },
+    more: {
+      label: '6. Care to share more?',
+      elementType: 'textarea',
+      elementConfig: {
+        placeholder: 'Anything more?...',
+        rows: 3
+      }
+    }
+  });
```
JuliaShavyrova: 1. this object is not updated during component lifecycle, so no need to use state here. please move this variable out of component body
2. you can use enum values instead of property names `more` => `[FeedbackFormElement.more]`

client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  return (
+    <S.FeedbackFormSection id="survey">
+      <Container>
+        <Flex directionColumn>
+          <h2>Fill Form Below</h2>
+          <form>
+            {formElementsArray.map(({ id, config }) => (
+              <S.FeedbackFormGroup key={id}>
+                <S.FeedbackFormLabel htmlFor="name">{config.label}</S.FeedbackFormLabel>
```
JuliaShavyrova: `htmlFor="name"` - looks like you need to pass `id` here

client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx
```diff
@@ -0,0 +1,124 @@
...
+                <Input
+                  elementType={config.elementType}
+                  id={id}
+                  elementConfig={config.elementConfig}
```
JuliaShavyrova: `Input` props can be named simplier, `type` and `config` names are pretty enough

client/src/components/Input/Input.styles.tsx
```diff
@@ -1,28 +1,90 @@
...
+
+  label::before {
+    left: -1.5px;
+    width: 1rem;
+    height: 1rem;
+    background-color: transparent;
+    border: ${({ themeType }) =>
+      themeType === 'light'
+        ? `1.75px solid ${theme.palette.grey.light}`
+        : `1.75px solid ${theme.palette.componentBackgroundDark}`};
```
JuliaShavyrova: can't you use ternary operator inside `1.75px solid ...` string?

client/src/components/Input/Input.tsx
```diff
@@ -1,79 +1,73 @@
...
+  radio: ({ id, elementConfig, theme, ...otherProps }: InputProps) => (
+    <S.InputGroup>
+      <Flex justifyFlexStart>
+        {elementConfig.options &&
+          Object.keys(elementConfig.options).map((option, i) => {
+            const values = elementConfig.options && Object.values(elementConfig.options);
+            return (
+              <S.RadioWrapper key={option} themeType={theme}>
+                <S.Radio id={option} name={id} {...elementConfig} {...otherProps} />
+                <label htmlFor={option}>{values && values[i]}</label>
```
JuliaShavyrova: `values && values[i]` => `elementConfig.options[option]`

client/src/components/Input/Input.tsx
```diff
@@ -1,79 +1,73 @@
...
+              </S.RadioWrapper>
+            );
+          })}
+      </Flex>
+    </S.InputGroup>
+  ),
+  rating: ({ id, elementConfig, theme, ...otherProps }: InputProps) => (
+    <StarRating
+      id={id}
+      numOfStars={elementConfig.options && Object.values(elementConfig.options).length}
```
JuliaShavyrova: `numOfStars` can be calculated inside `StarRating` component

client/src/components/StarRating/StarRating.tsx
```diff
@@ -0,0 +1,62 @@
...
+  ...otherProps
+}) => {
+  const [rating, setRating] = useState<number | null>(null);
+  const [ratingHovered, setRatingHovered] = useState<number | null>(null);
+
+  return (
+    <S.StarRating>
+      <Flex justifyFlexStart>
+        {[...Array(numOfStars)].map((_, i) => {
+          const ratingValue = i + 1;
```
JuliaShavyrova: why do you calculate rating value when it was already provided in config?

client/src/components/StarRating/StarRating.tsx
```diff
@@ -0,0 +1,62 @@
...
+          return (
+            <S.StarWrapper
+              key={`star-${i + 1}`}
```
JuliaShavyrova: you can use rating value instead of i+1

client/src/components/StarRating/StarRating.tsx
```diff
@@ -0,0 +1,62 @@
...
+              themeType={theme}
+              active={
+                !!(
+                  (ratingHovered && ratingValue <= ratingHovered) ||
+                  (rating && ratingValue <= (ratingHovered || rating))
```
JuliaShavyrova: please refactor condition, it can be simplier

client/src/components/StarRating/StarRating.tsx
```diff
@@ -0,0 +1,62 @@
...
+                )
+              }
+            >
+              <label htmlFor={id}>
+                <S.Star
+                  name={id}
+                  {...elementConfig}
+                  onClick={() => setRating && setRating(ratingValue)}
+                  onMouseEnter={() => setRatingHovered && setRatingHovered(ratingValue)}
+                  onMouseLeave={() => setRatingHovered && setRatingHovered(null)}
```
JuliaShavyrova: why do you need `setRating &&` and `setRatingHovered && ` checks?

client/src/types/index.ts
```diff
@@ -97,3 +99,23 @@ export type Theme = 'light' | 'dark';
...
+export type InputElement = 'input' | 'textarea' | 'radio' | 'rating';
+
+export interface InputProps {
+  elementType: InputElement;
+  id?: string;
+  elementConfig: {
+    type?: HTMLInputTypeAttribute;
+    placeholder?: string;
+    rows?: number;
+    options?: Record<string, unknown>;
```
JuliaShavyrova: `unknown` can be replaced to `string | number`

client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx
ViliusGaizutis: 1. I will expand it with input value, valid state, etc. which will be stored in this object.

client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx
ViliusGaizutis: What's wrong with what I currently have? I think it's pretty self-explanatory, if you look into an interface `InputProps`: 
1. `elementType` stands for custom type defined in type `InputElement` ('input' | 'textarea' | 'radio' | 'rating'), and I need this to be differently named from regular `type` attr to avoid misunderstandings.
2. `elementConfig` encompasses basic input html attributes like `type`, `placeholder`, `rows` etc. and maybe some custom as well.

client/src/components/Input/Input.styles.tsx
ViliusGaizutis: I can, thanks.

client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx
JuliaShavyrova: About `elementType` - sounds good, i understand now the logic. 
About `elementConfig` - your component doesn't have another configs, so no need to add additional word. In case this config is specified for certain part of component, you need to use more appropriate word instead of `element` (for example - `inputConfig`). If you're storing inside this object native input props only, you can name it `inputProps` as in mui library (https://mui.com/material-ui/api/input/)

ViliusGaizutis: `inputConfig` has not only native props, since `options` is a custom one. Which variant would look better?:
1. `inputType` and `inputConfig`
2. `elementType` and `config`
or maybe some other?

JuliaShavyrova: both options look good :) i'd recommend to use first option

ViliusGaizutis: Ok, I like it too:)

client/src/components/StarRating/StarRating.tsx
ViliusGaizutis: Sorry, it's a leftover from when I had code relating to star-rating in `Input` component and not in a separate component as it is now.

client/src/components/Input/Input.tsx
```diff
@@ -15,56 +15,47 @@ import { Flex } from 'styles/global';
...
-  rating: ({ id, elementConfig, theme, ...otherProps }: InputProps) => (
-    <StarRating
-      id={id}
-      numOfStars={elementConfig.options && Object.values(elementConfig.options).length}
-      theme={theme}
-      elementConfig={elementConfig}
-      {...otherProps}
-    />
+  rating: ({ id, inputConfig, theme, ...otherProps }: InputProps) => (
+    <StarRating id={id} theme={theme} inputConfig={inputConfig} {...otherProps} />
```
JuliaShavyrova: looks like now you can pass all props with spread operator, not `otherProps` only

client/src/components/StarRating/StarRating.tsx
```diff
@@ -10,43 +10,28 @@ import { IconStar } from 'assets/images/svg';
...
   const [rating, setRating] = useState<number | null>(null);
   const [ratingHovered, setRatingHovered] = useState<number | null>(null);
 
   return (
     <S.StarRating>
       <Flex justifyFlexStart>
-        {[...Array(numOfStars)].map((_, i) => {
-          const ratingValue = i + 1;
+        {[...Array(Object.values(inputConfig.options!).length)].map((_, i) => {
+          const ratingValue = Object.values(inputConfig.options!)[i] as number;
```
JuliaShavyrova: please define `Object.values(inputConfig.options!)` as variable

client/src/components/StarRating/StarRating.tsx
```diff
@@ -10,43 +10,28 @@ import { IconStar } from 'assets/images/svg';
...
-              key={`star-${i + 1}`}
+              key={`star-${ratingValue}`}
               themeType={theme}
-              active={
-                !!(
-                  (ratingHovered && ratingValue <= ratingHovered) ||
-                  (rating && ratingValue <= (ratingHovered || rating))
-                )
-              }
+              active={!!(ratingValue <= (ratingHovered! || rating!))}
```
JuliaShavyrova: operator `<=` always returns boolean, no need to add `!!`

client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx
JuliaShavyrova: when you'll start adding input values, just create another state (it can be with the same structure` {<input name>: <input value> }`). state should be used for dynamic values only, hardcoded text should be stored as separate variable

client/src/components/StarRating/StarRating.tsx
```diff
@@ -0,0 +1,49 @@
...
+const StarRating: FunctionComponent<InputProps> = ({ id, theme, inputConfig, ...otherProps }) => {
+  const [rating, setRating] = useState<number | null>(null);
+  const [ratingHovered, setRatingHovered] = useState<number | null>(null);
+
+  const ratings = Object.values(inputConfig.options!);
+
+  return (
+    <S.StarRating>
+      <Flex justifyFlexStart>
+        {[...Array(ratings.length)].map((_, i) => {
```
JuliaShavyrova: why not `ratings.map()`?

client/src/components/StarRating/StarRating.tsx
```diff
@@ -19,28 +19,25 @@ const StarRating: FunctionComponent<InputProps> = ({ id, theme, inputConfig, ...
...
+            key={`star-${ratingValue}`}
+            themeType={theme}
+            active={ratingValue <= (ratingHovered! || rating!)}
+          >
+            <label htmlFor={id}>
+              <S.Star
+                name={id}
+                {...inputConfig}
+                onClick={() => setRating(ratingValue as number)}
+                onMouseEnter={() => setRatingHovered(ratingValue as number)}
```
JuliaShavyrova: just interesting - adding type for `ratings` won't  allow to remove `as number`?
`const ratings: number[] = ...`
or even like this (difference will be that in this case `as` keyword will be used only once):
`const ratings = ... as number[]`

JuliaShavyrova (approved):

client/src/components/StarRating/StarRating.tsx
ViliusGaizutis: Originally ratings is of type `string | number[]`. If I try to set it as `const ratings: number[] = ...`, I get such error: "Type '(string | number)[]' is not assignable to type 'number[]'. Type 'string | number' is not assignable to type 'number'. Type 'string' is not assignable to type 'number'."

But yes, second option works.

PR #318, opened 2022-10-04
Change view for hourly weather

AndrewHelix:
Screenshot 2022-10-04 105919


src/components/main/HourlySection.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { HourlyWeatherType } from 'types/weatherTypes';
+
+type HourlySectionProps = {
+  weather: HourlyWeatherType[];
+};
+
+const getHourLabels = (weather: HourlyWeatherType[]): string[] => {
+  const labels: string[] = [];
+
+  for (let i = 0; i < weather.length; i++) {
```
JuliaShavyrova: why `for` instead of `forEach`?

src/components/main/HourlySection.tsx
```diff
@@ -0,0 +1,67 @@
...
+    () =>
+      weather.reduce((acc: { [index: string]: number }, el, i) => {
+        acc[i] = +convertTime(el.time).hours;
+        return acc;
+      }, {}),
+    [weather]
+  );
+
+  const currentWeatherByHour = weather.find(
+    el => +convertTime(el.time).hours === hoursSchema[hour]
```
JuliaShavyrova: please add type for `el` in methods callbacks

src/components/main/HourlySection.tsx
```diff
@@ -0,0 +1,67 @@
...
+  );
+
+  const hourLabels = useMemo(() => getHourLabels(weather), [weather]);
+
+  return (
+    <div className={styles['section-hourly']}>
+      {currentWeatherByHour && (
+        <>
+          <MainHourlyCard hourlyWeather={currentWeatherByHour!} />
+          <input
```
JuliaShavyrova: should range input be hidden if no current weather data provided? as far as i see, it's used in `MainHourlyCard` component only

AndrewHelix: No, I forgot delete this when move this code from Main component

JuliaShavyrova (approved):

src/components/main/HourlySection.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+type HourlySectionProps = {
+  weather: HourlyWeatherType[];
+};
+
+const getHourLabels = (weather: HourlyWeatherType[]): string[] => {
+  const labels: string[] = [];
+
+  weather.forEach((el: HourlyWeatherType, i: number) => {
+    if (i === 0 || i === 7 || i === 15 || i === 23) {
```
JuliaShavyrova, line 15: i'd recommend to use hours value for this condition (in case api will return disordered array), looks good for now. also you can use here array methods chaining to make this function look shorter and more declarative way

AndrewHelix: But I don't now smth about hours. Hours can start from 10 or 1 or 23. This condition adapted for any variant. 
You are right. If I get disordered array, I will have a problem because in UI I'll have values like '17.00 11.00 01.00 15.00'
In this case I only need to add sort method by hours before start function `getHourLabels`

PR #321, opened 2022-10-05
Added form validation, page-reload-check and some small UI changes for better UX

ViliusGaizutis:
Screenshot 2022-10-05 174745
Preview of form validation and page-reload-check.

Revisions:
1. [Added feedback form validation](https://github.com/ViliusGaizutis/ReactTask/commit/a8e36a952cebb04ccab3739e51a7626b16901a8a)
2. [Added page-reload-check and some small UI changes for better UX](https://github.com/ViliusGaizutis/ReactTask/commit/28ca682c74d017a27098e46ec76bebec9a67695c)


client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx
```diff
@@ -21,28 +34,37 @@ enum FeedbackFormInput {
 
 type FeedbackForm = {
```
JuliaShavyrova: small note - you can also define `FeedbackForm` using `Record` type: `Record<FeedbackFormInput, {...}>`

client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx
```diff
@@ -53,63 +75,209 @@ const feedbackFormConfig: FeedbackForm = {
...
+}));
+
 const FeedbackFormSection: FunctionComponent = () => {
-  const formElementsArray = Object.entries(feedbackFormConfig).map(entry => ({
-    id: entry[0],
-    config: entry[1]
-  }));
+  // State
+  const [feedbackForm, setFeedbackForm] = useState<FeedbackForm>({
+    [FeedbackFormInput.name]: {
```
JuliaShavyrova: please create `FeedbackFormSection.utils.ts` and move config and initial state object there

client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx
```diff
@@ -53,63 +75,209 @@ const feedbackFormConfig: FeedbackForm = {
...
+    }
+  };
+
+  useEffect(() => {
+    setIsFormDirty(feedbackFormInputs.some(input => input.value));
+  }, [feedbackFormInputs]);
+
+  useEffect(() => {
+    setIsFormValid(!feedbackFormInputs.map(input => input.valid).includes(false));
+  }, [feedbackFormInputs]);
```
JuliaShavyrova: useEffects callbacks can be combined to one function

client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx
```diff
@@ -53,63 +75,209 @@ const feedbackFormConfig: FeedbackForm = {
...
+    handleFormChange();
+
+    if (isFormValid) {
+      // I will write code here in another feature
+    }
+  };
+
+  useEffect(() => {
+    setIsFormDirty(feedbackFormInputs.some(input => input.value));
+  }, [feedbackFormInputs]);
```
JuliaShavyrova, line 89: `feedbackFormInputs` should be wrapped with useMemo, otherwise this effect will be called after every re-render

client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx
```diff
@@ -53,63 +75,209 @@ const feedbackFormConfig: FeedbackForm = {
...
+                  }`}</S.FeedbackFormLabel>
+                  <Input
+                    inputType={config.inputType}
+                    id={id}
+                    name={id}
+                    inputConfig={config.inputConfig}
+                    onChange={handleInputChange}
+                  />
+                  <S.FeedbackFormError>
+                    {feedbackForm[id as FeedbackFormInput].errMsg}
```
JuliaShavyrova, line 117: please avoid `as`

client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx
```diff
@@ -53,63 +75,209 @@ const feedbackFormConfig: FeedbackForm = {
...
+        [inputNames[currentIndex]]: {
+          ...form[inputNames[currentIndex]],
+          ...currField
+        }
+      }),
+      {}
+    );
+
+  const handleFormChange = () => {
+    const inputNames = Object.keys(FeedbackFormInput).map(key => key as FeedbackFormInput);
```
JuliaShavyrova: `map` usage is redundant, please remove. there is no need to extra use javascript methods for defining typescript types

client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx
```diff
@@ -53,63 +75,209 @@ const feedbackFormConfig: FeedbackForm = {
...
+
+    const validatedInputs = feedbackFormInputs.map((input, i) => {
+      const { isValid: valid, message: errMsg } = inputValidator(
+        input.value!,
```
JuliaShavyrova: please make `inputValue` param optional

client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx
```diff
@@ -53,63 +75,209 @@ const feedbackFormConfig: FeedbackForm = {
...
+      [inputName]: {
+        ...feedbackForm[inputName],
+        value: inputValue,
+        valid: isValid,
+        errMsg: message
+      }
+    });
+  };
+
+  const createUpdatedForm = (
```
JuliaShavyrova: 1. looks like this function also can be moved outside component body to component utils file
2. please don't forget to define return type

client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx
```diff
@@ -53,63 +75,209 @@ const feedbackFormConfig: FeedbackForm = {
...
+  config: entry[1]
+}));
+
 const FeedbackFormSection: FunctionComponent = () => {
-  const formElementsArray = Object.entries(feedbackFormConfig).map(entry => ({
-    id: entry[0],
-    config: entry[1]
-  }));
+  // State
+  const [feedbackForm, setFeedbackForm] = useState<FeedbackForm>({
```
JuliaShavyrova: please create 2 different interfaces for form state and form config

client/src/constants/index.ts
```diff
@@ -1 +1,22 @@
+import { InputValidator } from 'types';
+
+import { capitalize } from 'utils/helpers';
+
 export const BASE_URL = 'http://localhost:3000';
+
+export const VALIDATOR: Record<string, (value: string, param: number | boolean) => boolean> = {
```
JuliaShavyrova: `string` in first argument of `Record` is too common, please replace with more specific type

client/src/constants/index.ts
```diff
@@ -1 +1,22 @@
...
+  [InputValidator.required]: value => value.length !== 0,
+  [InputValidator.minLength]: (value, param) => value.length >= param,
+  [InputValidator.maxLength]: (value, param) => value.length <= param
+};
+
+export const VALIDATION_ERRORS: Record<
+  string,
```
JuliaShavyrova: the same here

client/src/utils/inputValidator.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  let isValid = true;
+  let message = '';
+
+  if (validation) {
+    const validators = Object.entries(validation);
+
+    isValid = !validators
+      .map(validator => {
+        const valid = VALIDATOR[validator[0]](value, validator[1]);
```
JuliaShavyrova: please define what is validator[0] and what is validator[1] using variables and types

client/src/utils/inputValidator.ts
```diff
@@ -0,0 +1,30 @@
...
+
+        const err = valid ? '' : VALIDATION_ERRORS[validator[0]](inputName, validator[1]);
+
+        message += ` ${err}`;
+
+        return valid;
+      })
+      .includes(false);
```
JuliaShavyrova: i don't like this mix of declarative and imperative ways if one function, you can do it better :)  you can create array using `validators.map` that contains objects `{ valid: boolean, message: string }` and then define common `message` and `valid` values using `map/join/includes` methods

client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx
ViliusGaizutis: I don't know how. If I leave it as `feedbackForm[id]`, I get compilation error "Element implicitly has an 'any' type because expression of type 'string' can't be used to index type 'FeedbackForm'. No index signature with a parameter of type 'string' was found on type 'FeedbackForm'." But I need this to be dynamic.

What about `{feedbackForm[FeedbackFormInput[id as keyof typeof FeedbackFormInput]].errMsg}`? (https://stackoverflow.com/questions/50417254/dynamically-access-enum-in-typescript-by-key)

client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx
ViliusGaizutis:
```
  const checkIsFormTouched = useCallback(() => {
    setIsFormDirty(feedbackFormInputs.some(input => input.value));
    setIsFormValid(!feedbackFormInputs.map(input => input.valid).includes(false));
  }, [feedbackFormInputs]);

  useEffect(() => {
    checkIsFormTouched();
  }, [checkIsFormTouched]);
```
Is the above ok?

JuliaShavyrova: looks good

client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx
JuliaShavyrova: please provide me exact code line/reason, why typescript thinks that `id` is a string, compilation error is not a reason to add `as` everywhere. yes, sometimes it's the only solution, but first you need to know why it was happened 

`{feedbackForm[FeedbackFormInput[id as keyof typeof FeedbackFormInput]].errMsg}` doesn't make code better

ViliusGaizutis:
```
const formElementsArray = Object.entries(feedbackFormConfig).map(entry => ({
  id: entry[0],
  config: entry[1]
}));
````
The above code makes typescript think `id` is a string, because `Object.entries()` converts enum keys into strings, so does `Object.keys()`. And in this case, `entry[0]` is a key of enum `FeedbackFormInput`. I can't seem to find any other workaround that does not involve using keyword `as`.

https://stackoverflow.com/questions/55012174/why-doesnt-object-keys-return-a-keyof-type-in-typescript
https://dev.to/prod/mastering-enums-in-typescript-1c1j

File: client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx (hunk 6 above)
Comment (ViliusGaizutis) [reply client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx]: It’s the same issue as in the comment above. I've used `Object.keys()` to get array of keys of enum `FeedbackFormInput` and had to set each key's type again. A better way I can think now is to define `type FeedbackFormInputKey = keyof typeof FeedbackFormInput`, as I am using this in several places. And in this particular case, I should write `Object.keys(FeedbackFormInput) as FeedbackFormInputKey[]`.

File: client/src/constants/index.ts (hunk 10 above)
Comment (ViliusGaizutis) [reply client/src/constants/index.ts]: This one again will involve using keyword `as`, as I can't think of any other way. I am thinking of defining a type `InputValidatorKey = keyof typeof InputValidator` and using it like so: `Record<InputValidatorKey, ...>`. Doing so would result in the line `const valid = VALIDATOR[validator[0]](value, validator[1]);` throwing the same error "Element implicitly has an 'any' type because expression of type 'string' can't be used to index type..." in the file `inputValidator.ts`, and the solution that I know would be to set `validator[0] as InputValidatorKey`, because a couple lines above I am again using `Object.entries()` like so: `const validators = Object.entries(validation);`.

File: client/src/utils/inputValidator.ts (hunk 12 above)
Comment (ViliusGaizutis) [reply client/src/utils/inputValidator.ts]:
```
const validationRule = validator[0] as InputValidatorKey;
const validationRuleValue = validator[1];
```
How about these?

File: client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx (hunk 5 above)
Comment (JuliaShavyrova) [reply client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx:117]: great! now i am sure that you see the full picture, not compilation error only :) `as` can be left here and for object.keys() too

File: client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx (hunk 6 above)
Comment (JuliaShavyrova) [reply client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx]: `Object.keys(FeedbackFormInput) as FeedbackFormInputKey[]` looks shorter, leave it

File: client/src/utils/inputValidator.ts (hunk 12 above)
Comment (JuliaShavyrova) [reply client/src/utils/inputValidator.ts]: good, but can be even better :) 
`.map(([name1, name2]) => {`

File: client/src/utils/inputValidator.ts [hunk 14]
```diff
@@ -1,30 +1,32 @@
...
+        const validated = VALIDATOR[ruleName as InputValidator](inputValue?.trim(), ruleValue);
+        const validationErr = VALIDATION_ERRORS[ruleName as InputValidator](inputName, ruleValue);
 
-        message += ` ${err}`;
+        return {
+          validatedList: [...validatedList, validated],
+          errMsg: `${errMsg} ${!validated ? validationErr : ''}`
+        };
+      },
+      { validatedList: [], errMsg: '' } as { validatedList: boolean[]; errMsg: string }
```
JuliaShavyrova: please define types for `reduce` method correctly, here is the doc:
`reduce<U>(callbackfn: (previousValue: U, currentValue: T, currentIndex: number, array: T[]) => U, initialValue: U): U;`

client/src/utils/inputValidator.ts
```diff
@@ -1,30 +1,32 @@
...
-        const err = valid ? '' : VALIDATION_ERRORS[validator[0]](inputName, validator[1]);
+    const result = validators.reduce(
+      ({ validatedList, errMsg }, [ruleName, ruleValue]) => {
+        const validated = VALIDATOR[ruleName as InputValidator](inputValue?.trim(), ruleValue);
+        const validationErr = VALIDATION_ERRORS[ruleName as InputValidator](inputName, ruleValue);
 
-        message += ` ${err}`;
+        return {
+          validatedList: [...validatedList, validated],
+          errMsg: `${errMsg} ${!validated ? validationErr : ''}`
```
JuliaShavyrova: please refactor `errMsg` to avoid extra spaces

client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx
```diff
@@ -190,39 +53,18 @@ const FeedbackFormSection: FunctionComponent = () => {
...
-        inputNames[i],
-        feedbackFormConfig[inputNames[i]].validation
-      );
-      return { valid, errMsg };
-    });
+    const validatedInputs = feedbackFormInputs.map((input, i) =>
+      inputValidator(input.value, inputNames[i], feedbackFormConfig[inputNames[i]].validation)
+    );
 
     const updatedForm = createUpdatedForm(validatedInputs, inputNames, feedbackForm);
```
JuliaShavyrova: `inputNames` variable is a constant that won't be updated, please move it to utils and use in `createUpdatedForm` function without passing as argument

client/src/utils/inputValidator.ts
JuliaShavyrova: 👆

ViliusGaizutis:
```
type ValidatorResult = { validatedList: boolean[]; errMsg: string[] };

const inputValidator = (
  inputValue: string,
  inputName: string,
  validation: Partial<InputValidation>
): { valid: boolean; errMsg: string } => {
  if (validation) {
    const validators = Object.entries(validation);

    const result = validators.reduce<ValidatorResult>(
      ({ validatedList, errMsg }, [ruleName, ruleValue]): ValidatorResult => {
        const validated = VALIDATOR[ruleName as InputValidator](inputValue?.trim(), ruleValue);
        const validationErr = VALIDATION_ERRORS[ruleName as InputValidator](inputName, ruleValue);

        return {
          validatedList: [...validatedList, validated],
          errMsg: [...errMsg, !validated ? validationErr : '']
        };
      },
      { validatedList: [true], errMsg: [''] }
    );

    return {
      valid: !result.validatedList.includes(false),
      errMsg: result.errMsg.join(' ')
    };
  }

  return { valid: true, errMsg: '' };
};
```
Is this better?

JuliaShavyrova: much better :)
i have a question - do you really need to have empty string and true here instead of empty arrays?
 `{ validatedList: [true], errMsg: [''] }`

ViliusGaizutis: I don't now, you're right:) Before I had to, it did not let me have it any other way.

PR #322, opened 2022-10-07
Added redux store, refactor main component, added air quality layer

AndrewHelix:
1) added layers toggle
Screenshot 2022-10-07 160605
2) refactor main component: move some parts to another new components
3) create air quality component. it use HourlySection and render necessary data (depend on layer)
Screenshot 2022-10-07 160552
4) added redux store for layers (air quality, forecasts) and theme
5) now if coordinates are in URL, position context will be update (if current position info in context differs from coordinates in URL )


src/components/main/AirQuality.tsx
```diff
@@ -0,0 +1,27 @@
...
+    state: { position }
+  } = useContext(positionContext);
+
+  useEffect(() => {
+    getWeather<{ forecast: AirQualityType[] }>(
+      '/air-quality/forecast/hourly/',
+      position.longitude,
+      position.latitude
+    ).then(res => serAirQuality(res.forecast));
+  }, [position.longitude, position.latitude]);
```
JuliaShavyrova: `[position]` should be enough

src/components/main/ForecastsDaily.tsx
```diff
@@ -0,0 +1,62 @@
...
+  } = useContext(positionContext);
+
+  const handleDaysSelected = (e: ChangeEvent<HTMLSelectElement>) => {
+    setSelectDays(e.target.value);
+    if (e.target.value) {
+      getWeather<{ forecast: DailyWeatherType[] }>(
+        '/forecast/daily/',
+        position.longitude,
+        position.latitude,
+        { periods: e.target.value, dataset: 'full' }
```
ViliusGaizutis: `e.target.value` can be placed in a const.

src/components/main/Main.tsx
```diff
@@ -1,134 +1,42 @@
...
-    }
-  };
+  const resultLongitude = +paramLongitude || position.longitude;
+  const resultLatitude = +paramLatitude || position.latitude;
 
   useEffect(() => {
-    if (resultLongitude && resultLongitude !== '0') {
-      fetchData(resultLongitude, resultLatitude);
+    if (resultLongitude !== position.longitude || resultLatitude !== position.latitude) {
+      changePosition(resultLatitude, resultLongitude);
```
ViliusGaizutis, line 31: You can shorten expression to `resultLongitude !== (position.longitude || position.latitude)`.

src/components/main/Main.tsx
```diff
@@ -1,134 +1,42 @@
...
-              </>
-            )}
-          </section>
-        </>
-      ) : (
-        <p>{positionError}</p>
-      )}
+      <LayersButton />
+      {layer === 'forecasts' && <Forecasts />}
+      {layer === 'air' && <AirQuality />}
```
ViliusGaizutis, line 39: `layer === 'forecasts'` and `layer === 'air'` could be defined as const or maybe as some sort of selector function `const isLayerForecasts = (layer: string) => layer === 'forecasts'`, since you use it in several places of your app. And then based on this you could write `isLayerForecasts ? <Forecasts /> : <AirQuality />`

src/store/theme/themeReducer.ts
```diff
@@ -0,0 +1,26 @@
...
+  document.body.dataset.theme = theme;
+  return theme;
+};
+
+export const themeReducer = (state: ThemeType = getTheme(), action: ThemeAction) => {
+  switch (action.type) {
+    case TOGGLE_THEME:
+      const theme = state === 'light' ? 'dark' : 'light';
+      setInLocalStorage(theme, THEME_LS_LABEL);
+      document.body.dataset.theme = theme;
```
ViliusGaizutis: You should think of another way to set item in `localStorage` and set `document.body.dataset.theme`. Reducers is not a place for that. Docs say "They must not do any asynchronous logic or other "side effects" (https://redux.js.org/tutorials/fundamentals/part-3-state-actions-reducers#writing-reducers). You could create an `actionCreator`.

src/components/main/Main.tsx
AndrewHelix: no. Longitude is longitude. I don't compare longitude with latitude

src/components/main/Main.tsx
AndrewHelix: I disagree with you. 
layer has union type (specific strings). And I can have 3 layers (also map, for example). In this case I can't use your solution `isLayerForecasts ? <Forecasts /> : <AirQuality />` and your function useless, if I want add the 3rd layer.

src/components/header/Header.tsx
```diff
@@ -23,8 +26,11 @@ const Header: FC = () => {
...
   const { changePosition } = useContext(positionContext);
-  const { theme, toggleTheme } = useContext(themeContext);
   const navigate = useNavigate();
+  const dispatch = useDispatch();
+  const { position } = usePosition();
+
+  const theme = useSelector(themeSelector);
 
   const getCities = async (search: string): Promise<{ locations: LocationInfoType[] }> => {
     const result = await HTTPRequest(`/api/v1/location/search/${search}`, {});
```
Auryte: Is it ok, to have this url here? Shouldn't it be in service or api files?

src/components/main/AirQuality.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import { getWeather } from 'services/getWeather';
+import { positionContext } from 'context/positionContext';
+
+import HourlySection from './HourlySection';
+
+import { AirQualityType } from 'types/airQualityType';
+
+const AirQuality: FC = () => {
+  const [airQuality, serAirQuality] = useState<AirQualityType[]>([]);
```
Auryte: typo :)

src/store/index.ts
```diff
@@ -0,0 +1,13 @@
...
+declare global {
+  interface Window {
+    __REDUX_DEVTOOLS_EXTENSION_COMPOSE__?: typeof compose;
+  }
+}
+
+export const store = createStore(
+  rootReducer,
+  window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ && window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__()
+);
```
Auryte, line 19: Could you explain, why do you need this window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE  ? And why do you put theme in dataset (document.body.dataset.theme)? (it's from reducer)

src/components/main/AirQuality.tsx
```diff
@@ -0,0 +1,27 @@
...
+    state: { position }
+  } = useContext(positionContext);
+
+  useEffect(() => {
+    getWeather<{ forecast: AirQualityType[] }>(
+      '/air-quality/forecast/hourly/',
+      position.longitude,
+      position.latitude
+    ).then(res => setAirQuality(res.forecast));
+  }, [position.longitude, position.latitude]);
```
JuliaShavyrova: `[position]` should be enough

src/components/main/AirQualityCard.tsx
```diff
@@ -0,0 +1,52 @@
...
+    AQI_SO2,
+    AQI_PM10,
+    AQI_PM2P5
+  } = airQuality;
+
+  const date = convertTime(time);
+
+  return (
+    <div className={styles['hourly-card']}>
+      <h4 className={styles['hourly-card__title']}>{`${date.hours}:${date.minutes}`}</h4>
```
JuliaShavyrova: i've already seen you have function `getTimeLabel`, please make this function shared and reuse everywhere

src/components/main/Forecasts.tsx
```diff
@@ -0,0 +1,81 @@
...
+      setCurrentWeather(res[1].current);
+      setHourlyWeather(res[2].forecast);
+    });
+  }
+
+  useEffect(() => {
+    if (position.longitude && position.latitude) {
+      fetchData(position.longitude, position.latitude);
+    }
+  }, [position.longitude, position.latitude]);
```
JuliaShavyrova: the same here about dependency

src/components/main/Forecasts.tsx
```diff
@@ -0,0 +1,81 @@
...
+import ForecastsHourly from './ForecastsHourly';
+import ForecastsDaily from './ForecastsDaily';
+import { positionContext } from 'context/positionContext';
+import { getCity } from 'services/getCity';
+import { getWeather } from 'services/getWeather';
+import { CurrentWeatherType, HourlyWeatherType } from 'types/weatherTypes';
+import { LocationInfoType } from 'types/cityInfoType';
+import { defaultLocationInfo } from './defaultStates';
+
+export type ViewType = 'cards' | 'graph';
```
JuliaShavyrova: please move this type to `types` folder

src/components/main/Forecasts.tsx
```diff
@@ -0,0 +1,81 @@
...
+              className={classNames(
+                styles['view-btn'],
+                view === 'graph' ? styles['active-view-btn'] : ''
+              )}
+              onClick={() => setView('graph')}
+            >
+              Graph view
+            </button>
+          </div>
+          <ForecastsHourly view={view} hourlyWeather={hourlyWeather} />
```
JuliaShavyrova: do you have ability to request hourly weather inside `ForecastsHourly` component?

src/components/main/HourlySection.tsx
```diff
@@ -52,7 +59,7 @@ const HourlySection: FC<HourlySectionProps> = ({ weather }) => {
   );
 
   const currentWeatherByHour = weather.find(
```
JuliaShavyrova: don't forget to add types

src/components/main/Main.css
```diff
@@ -100,13 +100,14 @@
...
 .hourly-input-range {
+  display: block;
   -webkit-appearance: none;
   width: 50%;
   height: 1rem;
   border-radius: 1rem;
   background: #d3d3d3;
   outline: none;
-  margin-bottom: 1rem;
+  margin: 0 auto 1rem auto;
```
JuliaShavyrova:
```suggestion
  margin: 0 auto 1rem;
```

src/components/main/Main.tsx
```diff
@@ -1,134 +1,42 @@
...
+import commonStyle from 'styles/commonStyles.css';
 
 const Main: FC = () => {
-  const [location, setLocation] = useState<LocationInfoType>(defaultLocationInfo);
-  const [currentWeather, setCurrentWeather] = useState<CurrentWeatherType | undefined>(undefined);
-  const [hourlyWeather, setHourlyWeather] = useState<HourlyWeatherType[]>([]);
-  const [dailyWeather, setDailyWeather] = useState<DailyWeatherType[]>([]);
-  const [selectDays, setSelectDays] = useState<string>('');
-  const [view, setView] = useState<'cards' | 'graph'>('cards');
+  const layer: layerState = useSelector(layerSelector);
```
JuliaShavyrova: don't forget to use PascalCase for type names

src/components/main/Main.tsx
```diff
@@ -1,134 +1,42 @@
...
+  const resultLatitude = +paramLatitude || position.latitude;
 
   useEffect(() => {
-    if (resultLongitude && resultLongitude !== '0') {
-      fetchData(resultLongitude, resultLatitude);
+    if (resultLongitude !== position.longitude || resultLatitude !== position.latitude) {
+      changePosition(resultLatitude, resultLongitude);
     }
-  }, [resultLongitude, resultLatitude]);
+  }, [resultLongitude, position.longitude, resultLatitude, position.latitude]);
```
JuliaShavyrova: 1.  `resultLongitude/resultLatitude` variables need to be wrapped with `useMemo`, otherwise this effect will be called every render
2. if `resultLongitude/resultLatitude` variables not used anywhere except this effect, move its declaration inside effect callback. the same for `paramLongitude/paramLatitude `

src/store/layer/layerReducer.ts
```diff
@@ -0,0 +1,18 @@
...
+interface ViewAction {
+  type: string;
+}
+
+export type layerState = 'forecasts' | 'air';
+
+export const layerReducer = (state: layerState = 'forecasts', action: ViewAction): layerState => {
+  switch (action.type) {
+    case TURN_ON_FORECASTS:
+      return (state = 'forecasts');
```
JuliaShavyrova: `return 'forecasts';` not working?

src/components/main/Main.tsx
AndrewHelix: 1. This component renders only when I open this page or when position is changed. In both cases component will render again. Of course for this two situations I need to know correct data about position. I think useMemo doesn't make sense in this case.
2. ok, I'll move this vars in useEffect callback. I think it'll solve this problem

src/store/index.ts
AndrewHelix: 1. I use this string for redux devtools https://github.com/reduxjs/redux-devtools/tree/main/extension#11-basic-store
2. I save theme in data attributes in `body` and use different css vars for different selectors: `body[data-theme='light']` and `body[data-theme='dark'] `

src/components/main/AirQuality.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const [airQuality, setAirQuality] = useState<AirQualityType[] | undefined>(undefined);
+  const {
+    state: { position }
+  } = useContext(positionContext);
+
+  useEffect(() => {
+    getWeather<{ forecast: AirQualityType[] }>(
+      '/air-quality/forecast/hourly/',
+      position.longitude,
+      position.latitude
```
JuliaShavyrova, line 20: next time please refactor `getWeather` function to get `position` object instead of `position.longitude, position.latitude`:  1. it will be shorter; 2. longtitude and latitude are related to each other and not used separately

JuliaShavyrova (approved):

PR #323, opened 2022-10-10
router pages for cities implemented

Silvik95:
1)Was added 'publicPath' to resolve the problem with reloading pages
image

2) Implemented new route "/city" to show the clicked city 
image

### DIFF
--- a/src/hooks/useFetch.js
+++ b/src/hooks/useFetch.js
@@ -1,13 +1,12 @@
 import { useEffect, useState } from 'react';
 import { API_EDPOIONTS, STORAGE_TOKEN } from '../helper/variables';
 
-function useFetch(endpoint, searchQuery, id) {
+function useFetch(endpoint, param) {
   const [isLoading, setIsLoading] = useState(false);
   const [error, setError] = useState(null);
   const [data, setData] = useState([]);
-  const URL = id
-    ? `http://localhost:3333/pfa.foreca.com/api/v1/${endpoint}/${id}`
-    : `http://localhost:3333/pfa.foreca.com/api/v1/location/${endpoint}/${searchQuery}`;
+  const URL = `http://localhost:3333/pfa.foreca.com/api/v1/${endpoint}/${param}`;
+
   useEffect(() => {
     (async () => {
       setIsLoading(true);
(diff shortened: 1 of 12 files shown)


Silvik95:
image
temperature added

Silvik95:
Now we can pass any id of the cities in our URL and it will work

src/routes/City.jsx
```diff
@@ -0,0 +1,34 @@
...
+      );
+      const reqDataName = await responseName.json();
+      setCityName(reqDataName.name);
+      const responseInfo = await fetch(
+        `http://localhost:3333/pfa.foreca.com/api/v1/current/${cityId}`
+      );
+      const reqDataInfo = await responseInfo.json();
+      setCityInfo(reqDataInfo.current);
+    })();
+  }, [cityId]);
```
virzhinia-petkova-redshelf: I'm wondering if you'll be able to use your query hook here

virzhinia-petkova-redshelf (commented):
Looks good :) One important comment

src/routes/City.jsx
Silvik95: image
that's how look  my hook logic

src/hooks/useFetch.js
```diff
@@ -1,20 +1,27 @@
...
+import { API_EDPOIONTS, STORAGE_TOKEN } from '../helper/variables';
 
-function useFetch(endpoint, searchQuery) {
+function useFetch(endpoint, searchQuery, id) {
   const [isLoading, setIsLoading] = useState(false);
   const [error, setError] = useState(null);
   const [data, setData] = useState([]);
+  const URL = id
+    ? `http://localhost:3333/pfa.foreca.com/api/v1/${endpoint}/${id}`
+    : `http://localhost:3333/pfa.foreca.com/api/v1/location/${endpoint}/${searchQuery}`;
```
virzhinia-petkova-redshelf: this is not a very extendable way to do it. Think about how you can just pass some params and directly update the url

src/hooks/useFetch.js
```diff
@@ -1,20 +1,27 @@
...
+        const response = await fetch(URL);
         const reqData = await response.json();
-        setData(reqData.locations);
+        if (endpoint === API_EDPOIONTS.SEARCH) {
+          setData(reqData.locations);
+        } else if (endpoint === API_EDPOIONTS.CURRENT) {
+          setData(reqData.current);
+        } else if (endpoint === API_EDPOIONTS.LOCATION) {
+          setData(reqData.name);
+        }
```
virzhinia-petkova-redshelf: what if you just the the data and get the property at the component level - won't it be a lot cleaner and without the endless conditions

virzhinia-petkova-redshelf (approved):
looks good

PR #326, opened 2022-10-12
Task 3 (Adding routing +about & error pages)

StefanAtanasov98:
Implementing routing in the app includes:

- Using react-router-dom;
- Adding about page with info about the app;
- Adding error page to show when a path is invalid;
- Adding navigation buttons to go from home to about and vice versa;

image


src/components/AboutPage/AboutPage.js
```diff
@@ -0,0 +1,22 @@
...
+import * as styles from '../../styles/AboutPage.module.css';
+import { useNavigate } from 'react-router-dom';
+
+function AboutPage(props) {
+  const navigate = useNavigate();
+
+  const aboutHeading = <h1>About us</h1>;
+  const aboutTxt = 'We have implemented this app to give access to people from all across the globe to the current and also future weather forecast for cities in any country. With the help of few clicks you can find information about the temperature, humidity, pressure, visibility, wind, as well as other weather characteristics for your chosen city of interest.';
+
+  const navigateToHome = () => navigate('/');
```
virzhinia-petkova-redshelf: do you think having the links in your header would be a better idea so you can navigate from each page to each other page

src/components/AboutPage/AboutPage.js
```diff
@@ -0,0 +1,22 @@
...
+
+  return (
+    <>
```
virzhinia-petkova-redshelf: here is a good example where a div instead of a fragment would probably be a better fit since you're wrapping an entire page

src/components/AboutPage/AboutPage.js
StefanAtanasov98: I thought about that and what made me not do it was the fact that for example on my home page I don't need a "Home" button and on my about page I don't need an "About" button. I guess I can still have them in my header but I thought this way of doing it is okay - the "Home Page" button is in a central spot on the about page so that it can't be missed.

virzhinia-petkova-redshelf: cool, but what about when you have more pages - are you going to add the links to all pages?

StefanAtanasov98: Yeah, you're right, I didn't think that far ahead :D So in this case when I'm on the about page and I click the "About" button (which will stay in the header),there should be no reloading, basically no action happening right?

src/App.jsx
```diff
@@ -1,18 +1,32 @@
...
 
 function App() {
   useAuthorize();
 
   return (
     <>
-      <Header />
-      <HomePage />
-      <Footer />
+      {/* <Header />
```
virzhinia-petkova-redshelf, line 15: delete comments if you don't need them

src/components/Homepage/Nav/Nav.js
```diff
@@ -1,16 +1,22 @@
 import React from 'react';
+import { useNavigate } from 'react-router-dom';
 import * as styles from '../../../styles/Nav.module.css';
 import { DropdownMenu } from './DropdownMenu.js';
 
 function Nav(props) {
+  // const navigate = useNavigate();
```
virzhinia-petkova-redshelf, line 7: delete what you don't need

virzhinia-petkova-redshelf (approved):
you can address the comments in your next PR

PR #327, opened 2022-10-12
Added functionality for form submitting and testimonials section

ViliusGaizutis:


ViliusGaizutis:
Screenshot 2022-10-12 000629
Screenshot 2022-10-12 000741
Preview of testimonials section.

Revisions:
1. [Added functionality for submitting feedback form and persisting feedb…](https://github.com/ViliusGaizutis/ReactTask/commit/fef5a97a705cb7316b7a3882acc91cccd0855d3c). Added function getFormData for getting data object for selected form state, new reducer for feedback state management and ability to persist state in localStorage.
2. [Added testimonials and made form inputs controlled](https://github.com/ViliusGaizutis/ReactTask/commit/f5a7f0455168cce8760f8cae430d8747637bee62). Added feedback testimonials section rendered in an accordion type component. Made form inputs controlled.
3. [Added some playful UI design elements](https://github.com/ViliusGaizutis/ReactTask/commit/79b413055e26f0310e32481eb9fafcebd24d670d). Added some svg blobs to make background and, UI in general, more playful.

client/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,45 @@
...
+import * as S from './Accordion.styles';
+
+interface AccordionProps {
+  index: number;
+  title: string;
+}
+
+const Accordion: FunctionComponent<AccordionProps> = ({ index, title, children }) => {
+  const [active, setActive] = useState(false);
+  const [isExpanded, setIsExpanded] = useState(index === 0);
```
JuliaShavyrova: 1. please add types
2. if both states are boolean, you can use the same rules about naming: `active/expanded` or `isActive/isExpanded`
3. in general, i'm a bit confused - what is the difference between these two states? i don't see code, where they are used separately

client/src/components/FeedbackPage/FeedbackFormSection/FeedbackFormSection.tsx
```diff
@@ -99,30 +116,36 @@ const FeedbackFormSection: FunctionComponent = () => {
...
-                    inputConfig={config.inputConfig}
-                    onChange={handleInputChange}
-                  />
-                  <S.FeedbackFormError>
-                    {feedbackForm[id as FeedbackFormInput].errMsg}
-                  </S.FeedbackFormError>
-                </S.FeedbackFormGroup>
-              ))}
+              {formElementsArray.map(({ id, config }, i) => {
+                const typedId = id as FeedbackFormInput;
```
JuliaShavyrova: please add type using  `as` for `formElementsArray` variable in case you need strict type for `id`, because now `typeId` variable exists only for type assignment

client/src/components/FeedbackPage/FeedbackTestimonialsSection/FeedbackTestimonialsSection.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+  const testimonialsArray =
+    testimonials &&
+    Object.entries(testimonials).map(entry => ({
+      id: entry[0],
+      content: entry[1]
+    }));
+
+  useEffect(() => {
+    const testimonialData = data.reduce<Testimonials>(
```
JuliaShavyrova: 1. please create separate function for generating testimonial data and move it to component `utils` file;
2. please create another config with type `Record<Testimonial, string>` for titles only to separate dynamic values and hardcoded strings

client/src/components/FeedbackPage/FeedbackTestimonialsSection/FeedbackTestimonialsSection.tsx
```diff
@@ -0,0 +1,159 @@
...
+                    .map(({ message, name, rating, timestamp }, i) => {
+                      const { day, year, time } = dateFormatter(timestamp, true);
+
+                      return (
+                        <S.FeedbackTestimonial key={`testimonial-${i + 1}`} themeType={theme}>
+                          <Flex directionColumn alignFlexStart>
+                            <S.FeedbackTestimonialTop>
+                              {rating && (
+                                <Flex justifyFlexStart>
+                                  {[...Array(5)].map((_, j) => (
```
JuliaShavyrova: i'd recommend to extend `StarRating` component functionality with "view" mode and reuse this component here

client/src/services/localStorage.ts
```diff
@@ -0,0 +1,22 @@
+export const loadState = () => {
+  try {
+    const serializedState = localStorage.getItem('state');
```
JuliaShavyrova: please move label "state" to variable

client/src/services/localStorage.ts
```diff
@@ -0,0 +1,22 @@
+export const loadState = () => {
```
JuliaShavyrova: don't forget to add types

client/src/utils/helpers.ts
```diff
@@ -3,3 +3,16 @@ export const capitalize = (string?: string): string =>
 
 export const getWeatherSymbol = (symbol?: string): string =>
   `https://developer.foreca.com/static/images/symbols/${symbol}.png`;
+
+export const throttle = (callback: any, limit: number) => {
```
JuliaShavyrova, line 7: why do you need this function?

client/src/components/Accordion/Accordion.tsx
ViliusGaizutis: 1. Sorry, I forgot. Will do.
3. Maybe I should've named second one differently? How about `isDefaultExpanded` or `isFirstExpanded`..? The idea is that you can toggle tabs active or inactive simultaneously, but I also wanted the first one to be active by default.

client/src/utils/helpers.ts
ViliusGaizutis: I've followed article https://medium.com/@jrcreencia/persisting-redux-state-to-local-storage-f81eb0b90e7e where it said that `saveState` should not be called too often "as it uses the expensive `JSON.stringify` operation", and it was suggested to use this utility function called `throttle` from `lodash` library, but I did not want to include library just for one function. So I found this simple function as a standalone.

Maybe it is redundant in this case, because `saveState` is called when feedback state changes, and this state is not updated that often, well, because you can't fill out forms that quickly :D But I think it would be nice to use this to persist `themeType`, and that could change often, as it only requires user to toggle switch-button.

client/src/components/FeedbackPage/FeedbackTestimonialsSection/FeedbackTestimonialsSection.tsx
ViliusGaizutis: I actually created type `Record<Testimonial, Record<'title', string>>` for the use in obj 
```
const testimonialsConfig: TestimonialsConfig = {
  [Testimonial.ratings]: {
    title: 'How would you rate our app?'
  },
  ...
};
```
to show that these strings stand specifically for title. I think it's nicer that way:)

client/src/components/FeedbackPage/FeedbackTestimonialsSection/FeedbackTestimonialsSection.tsx
ViliusGaizutis: Could you explain more by what you mean "view mode"? And is this really necessary? Yes, code structures do resemble each other a bit, for instance this
```
<S.FeedbackTestimonialStarWrapper
    key={`star-${j + 1}`}
    active={rating >= j + 1}
    themeType={theme}
  >
    <IconStar />
  </S.FeedbackTestimonialStarWrapper>
```
looks similar to
```
<S.StarWrapper
    key={`star-${ratingValue}`}
    themeType={theme}
    active={ratingValue <= (ratingHovered! || rating!)}
  >
    <label htmlFor={id}>
      <S.Star
        name={id}
        value={ratingValue}
        checked={value === ratingValue}
        {...inputConfig}
        onClick={() => setRating(ratingValue)}
        onMouseEnter={() => setRatingHovered(ratingValue)}
        onMouseLeave={() => setRatingHovered(null)}
        {...otherProps}
      />
      <IconStar />
    </label>
  </S.StarWrapper>
````
but just by a little. And what concerns styles, I've reused them by creating a mixin `starWrapperRoot`.

File: client/src/utils/helpers.ts (hunk 7 above)
Comment (JuliaShavyrova) [reply client/src/utils/helpers.ts:7]: ok, got your idea :) looks good, on real project you can also use throttle function imported from lodash library

File: client/src/components/Accordion/Accordion.tsx (hunk 1 above)
Comment (JuliaShavyrova) [reply client/src/components/Accordion/Accordion.tsx]: should first item be expanded only at start? in this case you can set default active state to true for first item and that's it

File: client/src/components/FeedbackPage/FeedbackTestimonialsSection/FeedbackTestimonialsSection.tsx [hunk 8]
```diff
@@ -101,8 +52,8 @@ const FeedbackTestimonialsSection: FunctionComponent = () => {
           <h2>Testimonials</h2>
           {data.length > 0 ? (
             <RequestDataWrapper data={data} loading={loading} error={error}>
-              {testimonialsArray?.map(({ id, content }, index) => (
-                <Accordion key={id} title={content.title} index={index}>
+              {testimonialsDataArray?.map(({ id, content }, index) => (
+                <Accordion key={id} title={testimonialsConfigArray[index].title} index={index}>
```
JuliaShavyrova: using `testimonialsConfigArray` you need always keep in mind that items ordering is very important. would be better to use `testimonialsConfig[id].title` instead

client/src/components/FeedbackPage/FeedbackTestimonialsSection/FeedbackTestimonialsSection.tsx
```diff
@@ -9,88 +9,39 @@ import RequestDataWrapper from 'components/RequestDataWrapper/RequestDataWrapper
...
+  title: entry[1].title
+}));
 
 const FeedbackTestimonialsSection: FunctionComponent = () => {
   const theme = useAppSelector(selectTheme);
   const { data, loading, error } = useAppSelector(state => state.feedback);
   const [testimonials, setTestimonials] = useState<Testimonials>();
 
-  const testimonialsArray =
+  const testimonialsDataArray =
```
JuliaShavyrova: please add types, we are still using typescript in project :)

client/src/components/FeedbackPage/FeedbackTestimonialsSection/FeedbackTestimonialsSection.utils.ts
```diff
@@ -0,0 +1,76 @@
...
+      rating?: number;
+      message: string;
+      timestamp: Date;
+    }[];
+  }
+>;
+
+export const generateTestimonialData = (data: Feedback[]) =>
+  data.reduce<Testimonials>(
+    ({ ratings, suggestions, recommendations, more }, curr): Testimonials => ({
```
JuliaShavyrova: `curr` can be destructured

client/src/components/FeedbackPage/FeedbackTestimonialsSection/FeedbackTestimonialsSection.tsx
JuliaShavyrova: i mean that would be better to extend StarRating component to have readonly "state", with non-clickable stars but showing provided value. yes, i see you have mixin for wrapper and reuse IconStar component, but these code snippets really look very similar + updating one component functionality/styles in one place will be easier than looking for multiple places in app + in future it would be easier to export this component with full functionality to another project

client/src/components/FeedbackPage/FeedbackTestimonialsSection/FeedbackTestimonialsSection.tsx
ViliusGaizutis: Ups, my bad. You're definitely right!:)

client/src/components/FeedbackPage/FeedbackTestimonialsSection/FeedbackTestimonialsSection.tsx
ViliusGaizutis: My bad again. It's just sometimes difficult to grasp where to place types...

client/src/components/Accordion/Accordion.tsx
ViliusGaizutis: OMG, how I did not see it.. I remember I tried to write condition in styles, like `transform: ${({ active, index }) => (index === 0 || active ? 'rotate(0deg)' : 'rotate(180deg)')};` and it stayed open on start, but no longer was togglable. Yeah, I went out of the roof with overcomplicating with this one :D

client/src/components/Accordion/Accordion.tsx
```diff
@@ -13,34 +13,20 @@ interface AccordionProps {
 }
 
 const Accordion: FunctionComponent<AccordionProps> = ({ index, title, children }) => {
-  const [active, setActive] = useState<boolean>(false);
-  const [isDefaultExpanded, setIsDefaultExpanded] = useState<boolean>(index === 0);
+  const [active, setActive] = useState<boolean>(index === 0 || false);
```
JuliaShavyrova: operator `===` always returns boolean, why do you need to add `|| false`?

client/src/components/FeedbackPage/FeedbackTestimonialsSection/FeedbackTestimonialsSection.tsx
```diff
@@ -52,29 +48,33 @@ const FeedbackTestimonialsSection: FunctionComponent = () => {
...
+                                  theme={theme}
+                                  inputConfig={{
+                                    type: 'radio',
+                                    options: {
+                                      'rating-1': 1,
+                                      'rating-2': 2,
+                                      'rating-3': 3,
+                                      'rating-4': 4,
+                                      'rating-5': 5
+                                    }
```
JuliaShavyrova: 1. looks like options for star rating component are always the same. can you please move them to separate variable and import in feedback form and testimonials component?
2. `readOnlyRating` can be renamed to `defaultValue` (it's more common and popular name for this prop)
https://reactjs.org/docs/uncontrolled-components.html#default-values

JuliaShavyrova (approved):

PR #328, opened 2022-10-12
Info page

Auryte:


client/src/API/get.ts
```diff
@@ -1,7 +1,19 @@
...
+  LocationByQuery,
+  LocationData
+} from 'types';
 
 const URL = 'http://localhost:5000';
+const headers = {
+  headers: {
+    Authorization: `Bearer ${localStorage.getItem('token')}`
+  }
+};
```
JuliaShavyrova: 1. please make object structure simplier, `headers` should not store object with only property `headers`
2. you can implement this functionality with adding auth header even better using axios interceptors
3. don't forget about types

client/src/components/ErrorPage/Error.tsx
```diff
@@ -0,0 +1,30 @@
...
+      <h2 className={styles.title}>Oh no...</h2>
+      <div className={styles.textBox}>
+        <p>
+          It looks like something has gone wrong, you may have mistyped the address or the page has
+          been removed.
+        </p>
+        <p>Try again or return to our homepage, where you can find all the latest news.</p>
+        <Link to="/">
+          <div className={styles.buttonBox}>
+            <Button type="submit" className={ButtonStyles.ErrorBtn}>
```
JuliaShavyrova: `type="submit"` is redundant, it's not a form

client/src/pages/Info/index.tsx
```diff
@@ -1,10 +1,101 @@
...
+import React, { FC, useRef } from 'react';
+import { CgDarkMode } from 'react-icons/cg';
+import { MdWbSunny } from 'react-icons/md';
+import { FaTemperatureLow } from 'react-icons/fa';
+import { getBannerImage } from 'utils/getImages';
+import styles from './styles.module.scss';
+
+interface Ref {
+  current: HTMLDivElement | null;
+}
```
JuliaShavyrova: you can use `RefObject<HTMLDivElement>` instead

client/src/pages/Info/index.tsx
```diff
@@ -1,10 +1,101 @@
...
+
+const Info: FC = () => {
+  const refSectionCurrent = useRef<HTMLInputElement | null>(null);
+  const refForecast = useRef<null | HTMLDivElement>(null);
+  const refTheme = useRef<null | HTMLDivElement>(null);
+
+  const handleClick = (ref: Ref) => {
+    ref.current?.scrollIntoView({ behavior: 'smooth' });
+  };
```
JuliaShavyrova: please try css solution first with anchors and smooth scrolling
https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior

client/src/utils/getImages.ts
```diff
@@ -8,3 +8,7 @@ export const getBackgroundImg = (url: string): string => {
     ''
   )}.jpg`;
 };
+
+export const getBannerImage = (): string => {
```
JuliaShavyrova: it should be a plain variable, not function

client/src/components/HeaderSearch/index.tsx
```diff
@@ -19,11 +19,11 @@ const HeaderSearch: FC = () => {
     }
   };
 
-  const handleSubmit = async event => {
+  const handleSubmit = async (event: { preventDefault: () => void }) => {
```
JuliaShavyrova: please use built-in react types instead

client/src/pages/Info/index.tsx
Auryte: I have tried, but it is not working

JuliaShavyrova: it should. you can ask Vilius for help (he used this solution) or ask me. also you can take a look on this conversation:
https://github.com/ximet/ReactTask/pull/305#discussion_r982563713

Auryte: oh, it's working, but only after second click. That's why I haven't noticed before.

client/src/pages/Info/index.tsx
```diff
@@ -45,16 +37,7 @@ const Info: FC = () => {
...
-                key={name}
-                href={`#${name}`}
-                onClick={event => {
-                  return handleClick(event, name);
-                }}
-                onKeyDown={event => {
-                  return handleClick(event, name);
-                }}
-              >
+              <a key={name} href={`about/#${name}`}>
```
JuliaShavyrova: `href={`#${name}`}` was correct

Auryte: No, this way I always go to the main page  http://localhost:9020/#theme
Screenshot 2022-10-14 154415

JuliaShavyrova: 1. url `<path>/#<anchor>` is anyway incorrect, anchors have url `<path>#<anchor>`
2. for future: in most cases it should work as defined first time, but exactly in your case it won't work - https://stackoverflow.com/a/46674310

client/src/pages/Info/index.tsx
JuliaShavyrova: it's not working correctly because you wrapped only small part of page with scroll-behavior property, but it should be placed in container that wraps both links and content

client/src/pages/Info/index.tsx
Auryte: It is working like this `href={`about#${name}`}`  :)  But somehow I can't get  smooth scroll behavior...

JuliaShavyrova:
also as far as i needed to go to your project to check smooth property, i have 2 more things that should be fixed: 
1. please fix error
image
2. rename screenshot images you're using for info page, assets names are also important

client/src/pages/Info/index.tsx
Auryte: I have tried to wrap all page, but scroll is the same both ways.

JuliaShavyrova: Please read conversation in Vilius PR one more time, take a look at his code, attentively check and define all required additional css properties.

Auryte: I want to ask about api get.ts. Is it worth reducing the code further? Perhaps it would be possible to reuse the try catch block, but I don't think that it would significantly affect the number of lines of code. What do you think?

JuliaShavyrova: get.ts file looks good for me, the only thing to refactor is types. i see in your function these return types and don't understand, why in one case you add ['data'] but in second one not. please provide reason or fix it
getHourlyWeather function: `Promise<AxiosResponse<HourlyWeather[]>['data'] | undefined>`
getDailyWeather function: `Promise<AxiosResponse<DailyWeather[]> | undefined>`

JuliaShavyrova (approved):

client/src/pages/Info/index.tsx
Auryte: I understand I have to use one method for consistency. I did it in different ways because I wanted you to tell me which way is better to use. Does it really matter?

JuliaShavyrova: i was not talking about methods, i was talking about types

Auryte: I did not express myself correctly. Yes I was talking about types. The different ways I type data/ forecast/current
 1 )` Promise< AxiosResponse<AuthenticationResponse>['data'] | undefined`; and then `const result: AxiosResponse<AuthenticationResponse> = `
 2) another way I create export type response types in types.ts - `
LocationDataResponse = Response & {
  data: LocationData; 
` and then ` const result: AxiosResponse<LocationDataResponse> =` 
}; 
 3) third way I get it like this : Promise<AxiosResponse<HourlyWeather[]>['data']  and then  ' const result = await forecaClient.get<{ forecast: HourlyWeather[] }>'

client/src/API/get.ts
```diff
@@ -1,84 +1,116 @@
...
+const forecaClient: AxiosInstance = axios.create({
+  baseURL: 'http://localhost:5000',
+  headers: {}
+});
 
-export const createToken = async () => {
+forecaClient.interceptors.request.use(
+  async (config: AxiosRequestConfig) => {
+    const accessToken: string | null | undefined = localStorage.getItem('token');
+    if (accessToken && accessToken !== 'undefined') {
```
JuliaShavyrova: please fix setting token in local storage (so it couldn't save 'undefined' string as value) and remove ` accessToken !== 'undefined'` condition

client/src/API/get.ts
```diff
@@ -1,84 +1,116 @@
...
+  }
+);
+
+export const createToken = async (): Promise<AuthenticationResponse> => {
   try {
-    const result = await axios.post(`http://localhost:5000/authorize/token?expire_hours=2`);
+    const result: AxiosResponse<AuthenticationResponse> = await axios.post(
+      `http://localhost:5000/authorize/token?expire_hours=2`
+    );
     window.localStorage.setItem('token', result.data.access_token);
```
JuliaShavyrova: also please create separate variable for 'token' label

PR #329, opened 2022-10-13
Added redux state for current weather

AndrewHelix:
Added redux state for current weather and in the future I'll add redux state for daily and hourly weather, and air quality

I have 1 moment. All works. I use useAppDispatch like in documentation. But I have mistake when I send Thunk action in dispatch function. I fix it by using type for dispatch function in component `const dispatch: CurrentWeatherThunkDispatch = useAppDispatch();`
If I will remove this type, I'll have mistake `Argument of type 'ThunkAction<void, CombinedState<{ layer: LayerState; theme: ThemeType; currentWeather: CurrentWeatherState; }>, unknown, CurrentWeatherAction>' is not assignable to parameter of type 'ViewAction | ThemeAction | CurrentWeatherAction'.` when send thunk action in dispatch `dispatch(loadCurrentWeather(position));`

Of course all works without mistakes if I use usual dispatch from useDispatch hook from react-redux library. But I think will be better if I use more types.

And my question for you: is my implementation normal or we have more preffered variant than mine?


src/store/currentWeather/currentWeatherReducer.ts
```diff
@@ -0,0 +1,59 @@
...
+
+interface CurrentWeatherState {
+  data: CurrentWeatherType | null;
+  error: string | null;
+  loading: boolean;
+}
+
+export interface RequestCurrentWeatherAction {
+  type: CurrentWeatherActions.REQUEST_CURRENT_WEATHER;
+  payload: null;
```
JuliaShavyrova: why do you need payload here?

src/store/hooks.ts
```diff
@@ -0,0 +1,6 @@
+import { TypedUseSelectorHook, useDispatch, useSelector } from 'react-redux';
+import { RootState } from './rootReducer';
+import { AppDispatch } from 'store';
+
+export const useAppSelector: TypedUseSelectorHook<RootState> = useSelector;
+export const useAppDispatch: () => AppDispatch = useDispatch;
```
JuliaShavyrova, line 7: have you tried this?
`export const useAppDispatch: () => CurrentWeatherThunkDispatch = useDispatch;`

src/store/currentWeather/currentWeatherReducer.ts
AndrewHelix: Maybe I thought remove prev state from redux state when I send new request)
I'll remove it, because I don't use payload data in reducer.
Of course I can add this logic in reducer, but I think more popular variant - only change loading property when we send new request)
ok?)

src/store/hooks.ts
AndrewHelix: Great! It is works!
But I'll create `export const useCurrentWeatherThunkDispatch: () => CurrentWeatherThunkDispatch = useDispatch;`
For each dispatch function which get her Thunk actions
ok?

JuliaShavyrova: "thunk" word is redundant, because every dispatch will be called with redux-thunk middleware :) you can replace it with 'App' or just leave `useCurrentWeatherDispatch`

src/store/currentWeather/currentWeatherReducer.ts
JuliaShavyrova: no need to add property in interface that won't be used (and in case `payload: null` it definitely won't be used) :)

JuliaShavyrova (approved):

PR #330, opened 2022-10-14
Added accessibility and some UX features

ViliusGaizutis:
Screenshot 2022-10-14 104551
Preview that shows spinner, custom scrollbar and tab focusing.

Revisions:
1. [Added spinner](https://github.com/ViliusGaizutis/ReactTask/commit/9ad4b65f48043b6a4b4c32254e73496de5786461)
2. [Improved accessibility according to WCAG](https://github.com/ViliusGaizutis/ReactTask/commit/5bebab9b42b822322ba50f3a66c7e40cf0f7a75a)
3. [Added custom scrollbar, removed noScrollbar mixin, used wrappers to w…](https://github.com/ViliusGaizutis/ReactTask/commit/4489cce4d2740058f16104a34a3055764aa1684e)
4. [Fixed small coloring issue dependent on theme type](https://github.com/ViliusGaizutis/ReactTask/commit/79068dfab1a7d97d162a36167782aa71d1c0b337)
5. [Changed styles dependent on user-select](https://github.com/ViliusGaizutis/ReactTask/commit/d6e4888406b3a39be68d1107a6ed280f52065054)
6. [Added small accessibility fixes](https://github.com/ViliusGaizutis/ReactTask/commit/5c65bb07d262cd22f6e6ddf4440c042bb87803a1)


client/src/components/AboutPage/AboutFeedbackSection/AboutFeedbackSection.tsx
```diff
@@ -19,7 +19,9 @@ const AboutFeedbackSection: FunctionComponent = () => (
         <h2>Leave A Feedback</h2>
         <p>Liked our service or maybe were dissatisfied with it?</p>
         <Link to="/feedback">
-          <Button color="white">Tell Us About It</Button>
+          <Button type="button" ariaLabel="Go to feedback page" color="white">
```
JuliaShavyrova: you can extend `ButtonProps` interface with `ButtonHTMLAttributes<HTMLButtonElement>` and define attributes the same way as for plain html button

client/src/components/Button/Button.tsx
```diff
@@ -4,11 +4,15 @@ import React, { FunctionComponent } from 'react';
...
 interface ButtonProps {
+  type?: 'button' | 'submit' | 'reset';
   color?: string;
+  ariaLabel: string;
 }
 
-const Button: FunctionComponent<ButtonProps> = ({ color, children }) => (
-  <S.Button color={color}>{children}</S.Button>
+const Button: FunctionComponent<ButtonProps> = ({ type, color, ariaLabel, children }) => (
+  <S.Button type={type} color={color} aria-label={ariaLabel}>
```
JuliaShavyrova: why button component doesn't contain any onClick handler?

client/src/components/ButtonSwitch/ButtonSwitch.tsx
```diff
@@ -11,20 +11,25 @@ interface ButtonSwitchProps {
   width: string;
   switchType: string;
   infoType?: string;
+  ariaLabel: string;
```
JuliaShavyrova: the same here about extending interface

client/src/components/Button/Button.tsx
ViliusGaizutis: I guess none of them has any. For instance, in a form I've put handler onto the form itself and not the submit button. I also have another button called ButtonSwitch, and that one has a handler, because I use it in two places, as a theme switch and as a forecast/air-quality switch.

client/src/components/ButtonSwitch/ButtonSwitch.tsx
ViliusGaizutis: But how would I get it from props? It's called "aria-label" and not "ariaLabel" (which is my custom prop).

JuliaShavyrova: you can provide it as 'aria-label', does it call issues? :)

client/src/components/Button/Button.tsx
JuliaShavyrova: i'd recommend to add onClick as far as it's a shared component. but anyway if you'll use recommended interface update, you can just write something like this:
<S.Button color={color} {...otherProps}>

ViliusGaizutis: You're right, using `{...otherProps}` did the job. I forgot about this way, even though I've used it in `Input` component. Although, eslint did not like it, so I had to turn off warnings/errors for custom components prop spreading.

JuliaShavyrova (approved):

PR #337, opened 2022-10-18
Task 5 (Adding a feedback page)

StefanAtanasov98:
Adding new page for user feedback includes:

- Adding 'FeedbackPage' component;
- Using local storage to save user input;
- Using redux to manage data;

image


src/components/FeedbackPage/FeedbackPage.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import * as styles from '../../styles/FeedbackPage.module.css';
+
+function FeedbackPage(props) {
```
virzhinia-petkova-redshelf, line 7: your feedback page supposed to save to local storage. Please take a look at the requirement again

StefanAtanasov98: Yes, I've made it so that when the user clicks the Submit button, the data is saved to local storage, I'm just trying to make redux work before pushing the code.

virzhinia-petkova-redshelf: ah, I thought you're ready with it. My bad/

src/components/Homepage/Homepage.js
```diff
@@ -3,6 +3,11 @@ import * as styles from '../../styles/HomePage.module.css';
 import { Nav } from './Nav/Nav.js';
 
 function HomePage(props) {
+  if (localStorage.getItem('userInput') === null) {
+    const userInput = {};
+    localStorage.setItem('userInput', JSON.stringify(userInput));
+  }
+
```
virzhinia-petkova-redshelf: I can't help but wonder why do you have this here

StefanAtanasov98: I thought that since the home page is the first thing that my app loads, it should be the page which creates the local storage object. But I guess if I put it in the feedback page it makes more sense?

virzhinia-petkova-redshelf: you can just create it when/if you need it. We put the auth logic when the app loads because you can't use the app without it. This is not the case here.

PR #340, opened 2022-10-19
Redux

Auryte:


client/.eslintrc.js
```diff
@@ -69,6 +69,7 @@ module.exports = {
     'import/no-extraneous-dependencies': ['error', { devDependencies: true }],
     'import/prefer-default-export': 'off',
     'no-unused-vars': 'warn',
+    'default-param-last': 'off',
```
JuliaShavyrova, line 72: why did you turn off this rule?

client/src/redux/actions.ts
```diff
@@ -0,0 +1,43 @@
...
+    try {
+      const feedback = getFromStorage(FEEDBACK_LOCALSTORAGE_LABEL);
+      if (feedback) {
+        const updatedFeedbackData = [newFeedback, ...feedback];
+        setInStorage(FEEDBACK_LOCALSTORAGE_LABEL, updatedFeedbackData);
+        dispatch({ type: FEEDBACK_REQUEST_SUCCESS, payload: { data: updatedFeedbackData } });
+      } else {
+        setInStorage(FEEDBACK_LOCALSTORAGE_LABEL, [newFeedback]);
+        dispatch({ type: FEEDBACK_REQUEST_SUCCESS, payload: { data: [newFeedback] } });
+      }
```
JuliaShavyrova: this part can be refactored, because in both cases you need to call `setInStorage` and `dispatch` functions with almost the same params. please rewrite it with no use if/else statement (ternary operator can be used)

client/.eslintrc.js
Auryte: Is it possible to switch state with action? I'm getting this error:
Screenshot 2022-10-19 184950

JuliaShavyrova: not sure this error is related to rule 'default-param-last'
UPD: i see the error, but not in the store.ts file. you can leave this rule, but please investigate and explain me where and why eslint showed you this error. error you provided above is not related to this rule

Auryte: It is this rule https://eslint.org/docs/latest/rules/default-param-last
Screenshot 2022-10-19 194655

client/src/components/Input/Input.tsx
```diff
@@ -15,6 +22,8 @@ const Input: FC<InputHTMLAttributes<HTMLInputElement> & InputProps> = ({
   value,
   onChange
 }) => {
+  const { darkMode } = useContext(DarkLightThemeContext);
```
JuliaShavyrova, line 25: it's not related to redux, next time please select files for pushing attentively

JuliaShavyrova (approved):

PR #343, opened 2022-10-22
Wrote tests

AndrewHelix:
Screenshot 2022-10-22 154553
Screenshot 2022-10-22 154417
Screenshot 2022-10-22 154501
 1) I didn't make tests for hourly weather thunk
2) I could not resolve problem with chart.js


src/mocks/fakeData.ts
```diff
@@ -0,0 +1,891 @@
...
+    precipRate: 0,
+    cloudiness: 100,
+    thunderProb: 0,
+    uvIndex: 0,
+    pressure: 1024.62,
+    visibility: 11243
+  }
+};
+
+export const fakeFeedbacks = [
```
JuliaShavyrova: i'd recommend to add type definitions to all mocked data so in case your types will be updated in future, you could easily update mock data

src/mocks/handlers.ts
```diff
@@ -0,0 +1,8 @@
+import { rest } from 'msw';
+import { fakeCurrentWeather, fakeHourlyWeather } from './fakeData';
+
+export const handlers = [
+  rest.get('https://pfa.foreca.com/api/v1/current/27.5373662,53.8972927', (req, res, ctx) => {
+    return res(ctx.json(fakeCurrentWeather));
+  })
```
JuliaShavyrova: do you still need it?

JuliaShavyrova (approved):

PR #345, opened 2022-10-24
Tests

Auryte:


Auryte:
Screenshot 2022-10-24 181030
Screenshot 2022-10-24 181049

client/src/components/ErrorPage/Error.spec.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  test('button is rendering', async () => {
+    render(<Button type="button" children="Return Home" className={ButtonStyles.ErrorBtn} />);
+    expect(screen.getByTestId(testIds.button)).toBeInTheDocument();
+  });
+
+  test('SecondaryPagesLayout is rendering', async () => {
+    render(<SecondaryPagesLayout children={undefined} />);
+    expect(screen.getByTestId(testIds.layout)).toBeInTheDocument();
+  });
```
JuliaShavyrova: in file 'Error.spec.tsx' you should add tests for Error component only, tests for Button, SecondaryPagesLayout and Main should be placed in there own files

client/src/components/ErrorPage/Error.spec.tsx
```diff
@@ -0,0 +1,46 @@
...
+    render(
+      <MemoryRouter>
+        <Routes>
+          <Route index element={<Main />} />
+        </Routes>
+      </MemoryRouter>
+    );
+    const links: HTMLAnchorElement[] = screen.getAllByRole('link');
+    expect(links[0].textContent).toEqual('Home');
+    expect(links[0].href).toContain('/');
```
JuliaShavyrova: please add test case for Error component where it should be checked `Return Home` button functionality

Auryte: but button has no functionality, it just renders. The main functionality covers Link element.

JuliaShavyrova: no matter, you're testing Error component, that can include anything inside. currently it includes some text and button (and i mean 'button' as UI element, no matter what components are used) with some functionality, so you should write a test to check this functionality. this test should emulate user click the button and check if it was a redirect to home page

client/src/components/ErrorPage/Error.spec.tsx
```diff
@@ -1,46 +1,45 @@
...
+    expect(link.getAttribute('href')).toBe('/');
   });
 
-  test('SecondaryPagesLayout is rendering', async () => {
-    render(<SecondaryPagesLayout children={undefined} />);
-    expect(screen.getByTestId(testIds.layout)).toBeInTheDocument();
-  });
-
-  test('User can navigate to a home page', async () => {
+  test('when button clicks, there is link to "/"', async () => {
```
Auryte: I did the tests according to info I found " jsdom is simulating a browser but it has some limitations. One of these limitations is the fact that you can't change the location. If you want to test that your link works I suggest to check the href attribute of your <a>.  Is it right?

JuliaShavyrova: you have client-side routing in your app, that's managed with react-router library (remember, you're wrapping the whole app with BrowserRouter/HashRouter or else component imported from react-router library). so you need to do the same in the test, like here:
https://stackoverflow.com/questions/70974339/testing-navigation-with-react-testing-library

JuliaShavyrova (approved):

PR #346, opened 2022-10-24
Task 6 (Implementing dark mode)

StefanAtanasov98:
Implementing dark mode includes:

- Creating dark mode css files;
- Managing state for dark mode flag;
- Designing a toggle button in header;

image

### DIFF
--- a/src/styles/colours.module.css
+++ b/src/styles/colours.module.css
@@ -1,11 +1,21 @@
 :root {
-  --lightBlue: #3498db;
-  --blue: #2980b9;
-  --darkBlue: rgb(2, 2, 117);
-  --darkerBlue: rgb(2, 2, 91);
-  --lightPurple: #b393d3;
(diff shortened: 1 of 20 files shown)


src/styles/colours.module.css
```diff
@@ -1,8 +1,11 @@
...
   --blue: #2980b9;
+  --darkBlue: rgb(2, 2, 117);
+  --darkerBlue: rgb(2, 2, 91);
   --lightPurple: #b393d3;
+  --darkPurple: rgb(75, 4, 75);
   --white: rgb(255, 255, 255);
-  --grey: #ddd;
+  --grey: #ccc;
   --black: rgba(0, 0, 0);
 }
```
virzhinia-petkova-redshelf: you have a set of variables. Can you overwrite them with different colors depending on the theme? The way you have it set up currently is not very extendable. What if you wanted to have more palettes - how easy would it be to add them?

PR #349, opened 2022-10-26
Search suggestions

Auryte:


client/src/API/localStorage.ts
```diff
@@ -1,6 +1,6 @@
 import { Feedback } from 'types';
 
-export const getFromStorage = (name: string): Feedback[] | null => {
+export const getFromStorage = (name: string): any => {
```
JuliaShavyrova: you can add generic here to define what type should be returned instead of any

client/src/components/HeaderSearch/index.tsx
```diff
@@ -1,61 +1,51 @@
...
   const [inputValue, setInputValue] = useState<string>('');
-  const navigate = useNavigate();
+  const [displaySearchResults, setDisplaySearchResults] = useState<boolean>(false);
 
-  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-    if (isInputValid(event.target.value)) {
-      setInputValue(event.target.value);
-    }
+  const handleInputFocus = (): void => {
+    return setDisplaySearchResults(true);
```
JuliaShavyrova: `void` means that function should not return anything, please remove `return`

client/src/components/HeaderSearch/index.tsx
```diff
@@ -1,61 +1,51 @@
...
-          `${locationData.locations[0].name}-${locationData.locations[0].country}`
-        );
-        navigate(`/details/${urlLocationName}`);
-      }
-    } else {
-      setStatusMsg('Please, enter the name of the city. The input cannot be empty');
-      setCoordinates(null);
-    }
+  const handleInputBlur = (): void => {
+    return setDisplaySearchResults(false);
```
JuliaShavyrova: same here

client/src/components/Loader/Loader.tsx
```diff
@@ -1,9 +1,16 @@
...
+  search = 'searchContainer',
+  main = 'loaderContainer'
+}
+interface LoaderProps {
+  className: string;
+}
+export const Loader: FC<LoaderProps> = ({ className }) => {
   return (
-    <div className={styles.loaderContainer}>
+    <div className={styles[className]}>
```
JuliaShavyrova: i'd recommend to move searchContainer style to the component that uses Loader and needs it to be customized. and pass this className using props. in this case your loader should look like this:
```suggestion
    <div className={`${styles.loaderContainer} ${className}`}>
```

client/src/components/SearchResultsList/SearchResultsList.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  return (
+    <ul className={`${styles.container} ${darkMode ? styles.dark : styles.light}`}>
+      {isLoading ? (
+        <Loader className={LoaderStyles.search} />
+      ) : (
+        searchData?.locations.slice(0, 10).map(({ lon, lat, id, name, country }) => {
+          return (
+            <li className={styles.listElement} key={id}>
+              <div
```
JuliaShavyrova: please use semantically correct tags or react-router components, divs usually are not clickable. also it will help you to avoid `as` here:
`const target = event.target as HTMLDivElement;`

client/src/components/SearchSavedLocations/SearchSavedLocations.tsx
```diff
@@ -0,0 +1,47 @@
...
+      lat: Number(lat),
+      lon: Number(lon)
+    };
+    return searchLocationCoords;
+  };
+  return (
+    <div>
+      {savedSearch && (
+        <li className={styles.listElement}>
+          <div
```
JuliaShavyrova: the same about div here

client/src/components/SearchResultsList/SearchResultsList.tsx
```diff
@@ -0,0 +1,77 @@
...
+        return (
+          <SearchSavedLocations onClick={handleClick} savedSearch={savedSearch} key={savedSearch} />
+        );
+      })}
+      {!prevSearches.length && (
+        <li className={styles.listElement}>
+          <div className={styles.locationName}>No saved locations</div>
+        </li>
+      )}
+    </ul>
```
JuliaShavyrova, line 72: use ternary operator instead of these 2 expressions:
`prevSearches?`, `!prevSearches.length &&`

client/src/pages/Details/index.tsx
```diff
@@ -21,15 +14,13 @@ const Details: React.FC = () => {
...
-        const searchLocationData = await getLocationByQuery(locationName);
-        if (searchLocationData) {
-          const searchLocationCoords = {
-            lat: searchLocationData?.locations[0].lat,
-            lon: searchLocationData?.locations[0].lon
-          };
-          setCoordinates(searchLocationCoords);
-        }
+        const lon = location.split('(')[1].split(')')[0];
+        const lat = location.split('(')[2].split(')')[0];
```
JuliaShavyrova: please create shared function for parsing saved search string and reuse it everywhere

client/src/redux/searchReducer.ts
```diff
@@ -0,0 +1,53 @@
+import { SEARCH_LOCALSTORAGE_LABEL } from 'constants/labels';
+import { AnyAction } from 'redux';
+import { SearchState } from 'types';
+import { getFromStorage, setInStorage } from '../API/localStorage';
+
+export const SEARCH_SAVE_CURRENT_LOCATION = 'SAVE_CURRENT_LOCATION';
```
JuliaShavyrova: where do you use this var? i don't see it inside reducer body

client/src/redux/searchReducer.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+    const prevSearches = [action.payload, ...state.prevSearches];
+    setInStorage(SEARCH_LOCALSTORAGE_LABEL, prevSearches);
+    return {
+      ...state,
+      prevSearches
+    };
+  }
+  return state;
+};
```
JuliaShavyrova, line 25: 1. reducer should be a pure function, so it's not recommended to add functions related to local storage here. you can use subscribe method to update local storage values
https://stackoverflow.com/a/37690899
2. i see 3 times almost the same code lines, that means the code should be refactored:
```
 const prevSearches = [action.payload, ...slicedPrevSearches];
setInStorage(SEARCH_LOCALSTORAGE_LABEL, prevSearches);
    return {
      ...state,
      prevSearches
    };

```

client/src/types.ts
```diff
@@ -109,9 +109,34 @@ export interface Feedback {
...
 export type Errors = Record<Validations, (value: string, param: number) => string>;
 
-export interface FeedbackState {
+export type FeedbackState = {
   feedback: Feedback[];
   loading?: boolean;
   error?: string;
+};
+export interface SearchState {
+  prevSearches: any;
```
JuliaShavyrova: please add type

client/src/redux/searchReducer.ts
Auryte: I solved this problem with action creator as I have already did before with feedback reducer. I will definitely read and learn about subscribe method in the near future. It is too late and not enough time for my tired head :)

